PR #16, opened 2014-08-02
tests

tokejepsen:
I originally did a quick test where the only tag on the nodes where a boolean "publish", then I got the export data types from what ever the type of the shape node were. This I found to be too workflow specfic.

Second iteration is specifying the data export type on the node itself.


mottosso:
Having some issues. We'll have to find some way of getting references to work when we aren't using the same root directories. I've had this issue too.

Relative paths work, but only when setting the project, and that won't work when setting the project for the rig, referencing the model, and then switching to a shot, referencing the rig; the rig won't know where to find the model.

Can you think of a way? Or maybe lets just not use referencing? That would be a shame.. Thoughts?

tokejepsen:
could not use reference for a little while, but eventually to test production scenarios we'll have to have use references. I'll have a think about it.

mottosso:
Mm, agreed. References would help a lot.

Maybe it's possible for Maya to accept an environment variable as root. E.g.

```
%ROOT%/assets/myasset.mb
```

And we'd all agree on one to use during dev. But I'm sceptical about the platform independence of it.

mottosso:
Ah, okay I see what you did. So essentially, each controller has:

``` json
{
  "publish": true,
  "dataExport": ["json", "atom"]
}
```

Some general notes:

What do you think about generally sticking  "publishable" as attribute for anything publishable? And what do you think about the concept of "classes", as opposed to specifying exact output types? E.g.  "model" or "animation" instead of "obj" and "json"? Classes would then be configured independently.

There hasn't been much talk about it so far, maybe we should start a thread about it, but this is where a summary would end up.
https://github.com/abstractfactory/publish/wiki/classes

tokejepsen:
I dont have a problem with the using "publishable", I usually just try to keep the names as short as possible to avoid extra typing.

Classes would be definitely be a better idea, good shout!

PR #17, opened 2014-08-02
Making scene publishable

mottosso:
Consolidated "sphere" character into single group, imported references and added an objectSet for publishing in the shot.

The scene is in a publishable state at this point, the only thing remaining is whether or not Publish can figure out what to ultimately export. Each control has an identifying attribute on it, so it should work.

Thoughts?


tokejepsen:
Im not entirely sure of the use of sets. Im guessing it is to provide the user with some feedback on what is getting published, and a way to manage the publish. I would say that we dont need the sets, and provide GUIs that provides the user with info and tool to manage the publish. Sets also seems like a workflow restriction.

The grouping of assets wouldn't be needed either as the tools could find the nodes to publish within the scene.

mottosso:
Sets are a way of storing settings in each scene. Its a concept I borrowed from Softimage, in which tools store their settings in an individual node per tool-session.

Essentially, the tool (Publish) could, as you say, handle selecting and provide feedback. It could then use sets as a means of storing any settings they make. The tool would then "remember" settings on a per-scene basis.

Until we've got a GUI however, we'll have to do it manually, so as to make our tests work.

Does that make sense? Feel free to suggest other means.

mottosso:
It brings up another question which is probably better suited for the Issues section; do we want a publishable scene to be batch-publishable?

We could then delegate publishing to another machine, such as for distributed publishing, amongst other good stuff.

tokejepsen:
I would maybe use something else for storing the tools settings, like a non-DAG node to hide it away from the user so they only interact through the GUI to minimize user errors. As the workflow is atm, there isn't a use for storing the settings, as we could just modify the attributes themselves which are then scene specific, but there will most likely be a point in the future where this is nessecary.

We definitely want batch publishing:)

mottosso:
Sure, we could do that. But, until we've got a GUI and unless we store settings, how do we publish? :)

mottosso:
Some settings examples:



This could potentially be publishable, without a GUI.

tokejepsen:
I would start building methods for getting the data, and modifying it. Again without using the sets, so its all manual coding inputs for now. When we are a comfortable level with the code blocks, we can start making GUIs. If you need to modify multiple items in a scene, it is up to a bit of scripting/hacking to get it done.

I was wondering about the storing of the start/end frame, which doesn't make sense to me as I would just tell Publish to export a range. Or if batch processing, looking at the timeline information.

tokejepsen:
I would be tempted as well to use 'network' nodes in Maya for storing the information instead of sets. They are relatively hidden away from the immediate user, but as a TD or for development you could get to them through the outliner (non-DAG nodes) or hypershade.

mottosso:
Sure, I don't mind any of that.

As for sets, I'll list some of the features of sets that I found useful in an issue, maybe there are better ways of getting to those than sets.

Closing this for now, feel free top open it back up if you have anything to add.

PR #22, opened 2014-08-05
Mockup of Publish for Maya with test

mottosso:
Getting the ball rolling, everything is currently up for discussion - modifications and comments are welcome.

To try this pull-request, pull directly from the @mottosso account. For details, see here:
- [Stackoverflow](http://stackoverflow.com/questions/4169832/update-my-github-repo-which-is-forked-out-from-another-project)
- https://help.github.com/articles/fetching-a-remote
## Behaviour
1. `class` was renamed to `family` to avoid conflict with Python's use of it
2. Program is a single file
3. There is a single mesh in the scene
4. The family used is `model`
5. Selection is done via objectSet, although I encourage trying out attribute tagging as well.
6. An additional selection is done to illustrate how playblasts could be published
## Usage

Publish is added to the File-menu, let's talk about whether or not this is a good idea.
1. Add publish/integration to your PYTHONPATH
2. Open Maya
3. Open [test](https://github.com/mottosso/publish/tree/master/resources/tests/selection/multiple_selection)
4. File --> Publish
5. Published file ends up in $MAYA_PROJECT_DIR/published/{date}/{date}.mb

If not run from the File menu, you can do:

``` python
import publish.maya as publish

# Convenience
publish.publish()
```

Or, for explicit use:

``` python
import publish.maya as publish

# parse selection
selection = publish.select()

# validate
failures = publish.validate(selection)

# extract
if not failures:
    for instance in selection:
        path = publish.extract(instance)
        print "Extracted {0}".format(path)

        # conform
        publish.conform(path)

else:
    log.info("There were errors:")
    for failure in failures:
        log.info("\t{0}".format(failure))

print "Successfully published scene"
```

If no project is set, a `published` directory is created next to the currently open file.
## Flawed
1. No gui, nor notifications of any kind, except for logging messages in script editor
2. Maya-only
3. There is no tagging or consideration of additional nodes other than objectSets
4. Code is in a single file
5. Only supports `model` family
6. Conform does nothing (@ljkart, this is where your template system could come in)
7. And lots more I'm sure.


mottosso:
It struck me that this version doesn't export to a temporary local directory first, it should!

PR #29, opened 2014-08-08
Refined plug-in system

mottosso:
## Plug-ins

Functional plug-in system, plug-ins are looked up at run-time and integrated into the validation process based on which family each instance belongs to.

E.g. an instance of type _"model"_ is assigned all plug-ins of type _"validator"_ for family _"model"_ for its supported _"host"_ (e.g. Maya).

Summarising its behaviour here:
- https://github.com/abstractfactory/publish/wiki/Plugin-system
## Tests

Also in this request are tests running without a fixture; e.g. a Maya scene is constructed within each test. 

The difference between this and the [existing tests](https://github.com/abstractfactory/publish/tree/master/resources/tests) is that these aren't used interactively within Maya, but rather triggered automatically by `python -m unittest discover`.

Have a look at [test_mayapy.py](https://github.com/mottosso/publish/blob/708b2e76733dead8b3d7a32e75ebecd99525e354/publish/tests/test_mayapy.py) for an example.
## Persistent configuration

An initial version of externalised configuration, in config.json. The idea is for users to modify this file and thus affect the overall operation of Publish.
- See also [config.json](https://github.com/mottosso/publish/blob/708b2e76733dead8b3d7a32e75ebecd99525e354/publish/config.json)


mottosso:
Just wanted to point out that this isn't how merges are supposed to happen (i.e. in the dark), they should always have a discussion first!

Apologies for this, I'll try and keep a tighter leash on git.

PR #30, opened 2014-08-08
First prototyping (wip) for modular design based on Actions.

BigRoy:
- Separate base classes into core.py
- Remove redundant code from maya.py for this design.
- Add examples for Selectors, Validators, Extractors.
- NOTE: There's no working prototype/code at this point (consider WIP)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 *.swatches
+.idea
--- a/publish/core.py
+++ b/publish/core.py
@@ -1,0 +1,53 @@
+
+
+# Base Classes
+# ============
+class Context(object):
+    pass
+
+
+class Action(object):
+    """ Action that operates on a predefined Context.
+        An action will have to define what type of Context it requires as well as
+        what data on such Context it uses. This way we can raise errors when one of
+        the dependencies is not met with a certain Context.
+
+        `__inputs__` refers to what data this Action requires to already exist on the Context
+        `__outputs__` refers to what data this Action will add/edit on the Context
+    """
+    __context__ = Context   # The type of Context the action operates on
+    __inputs__ = None
+    __outputs__ = None
+
+    def __init__(self, context=None):
+        self.__context = None
+
+        if context is not None:
+            self.context = context
+
+    @context.setter
+    def context(self, c):
+        assert isinstance(c, self.__context__)
+        self.__context = c
+
+    @property
+    def context(self):
+        return self.__context
+
+
+# The different types of default Action that operate on a Context: Selector, Validator, Extractor
+class Selector(Action):
+    def process(self, context=None):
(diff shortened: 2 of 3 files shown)


mottosso:
Excellent, having a look now!

BigRoy:
Something to note is that the _Actions_ are now directly interacting with the _Context_ in `Action.process()`. This allows the user to accidentally change the _Context_ somewhere the _Action_ should not. Instead the _Action_ should only be allowed access to its inputs/outputs. This way the _Action_ also doesn't need to know the parameters name on the _Context_ but only on itself, which is great because that's what the user is actually implementing.

Think of the inputs and outputs as _Attributes_ on a _Node_ where the Node is the Action. The Action doesn't have access to anything other than the inputs it requested. _To avoid confusion: The Action CAN access stuff in a Maya scene or something in a folder on you HDD, but the input of the Action will configure what objects it will look for or what the folder's path is._

Let me know if this doesn't make any sense.

mottosso:
How do you feel about renaming "Action" -> "Filter" to align with [Pipes and Filters](http://msdn.microsoft.com/en-us/library/ff647419.aspx)?

BigRoy:
Personally I think _Filter_ sounds confusing and _Action_ seems more accurate. I also thought about calling it _Node_ because that might be more familiar for people in our industry, especially if we go with a more _Node Graph_ style. And _Filter_ might be confusing with Selectors that 'filter' the provided _Context_. I'm currently also thinking about how the Artist perceives what something is and not necessarily the TD/Programmer that might recognize it from a data flow architecture.

Nevertheless I think I could also get used to _Filter_. Feel free to adapt it and lets see where that leads us?

mottosso:
>  I'm currently also thinking about how the Artist perceives what something is and not necessarily the TD/Programmer that might recognize it from a data flow architecture.

Our domain language doesn't necessarily have to influence the user interface. I'm all for calling objects in a visual node graph "nodes" as opposed to "filters", like they are in Maya and Nuke et. al. Using "filter" in our code-base however would make for a quicker introduction to newcomers already familiar with the concept along with keeping our heads on the same page as we'd be looking at the same references for inspiration and understanding.

Additionally, "action" is already reserved terminology within the [Command Pattern](http://en.wikipedia.org/wiki/Command_pattern) which we may use within the GUI.

BigRoy:
That's correct! 
Let's go with _Filter_!

Not sure about disconnecting between code and UI though, but we'll see when we get there!

> **Artist:** "Man... that Action here is giving me a hard time!"
> **TD:** "Oh yeah, let me update that Filter."
> _A little later._
> **Artist:** "These programmers are so weird!"
> **Artist2:** "Tell me about it."

PR #35, opened 2014-08-17
Initial commit of Selectors

tokejepsen:
This is a direct piggy back from the mottosso pull request #34. I'm still getting my bearings on this project, but this should implement Selectors.

It is just taking the workflow from the validators/extractors, and putting it into selectors. Only difference is that the base class is not Filter, but a new base class called Selector.

Let me know what went wrong:)


BigRoy:
What's the reason for choosing to inherit Validators/Extractors from Selectors? Why do you think the Filter isn't the responsible Base class in this regard?

tokejepsen:
The Validators/Extractors are still inheriting from the Filter base class, but since the Filter class gets initialized with an instance, my poor brain couldnt think of a way to use the Filter as a base class for the Selector hence the new base class. On second thought I could maybe have provided a default for the instance instead.
Though there is something that I found limiting about the plugin module, which was its discover function that used string patterns of the plugin module name, to categorize the plugin. Maybe we should change that to look at which class (Validator/Extractor/Selector) it is coming from?

mottosso:
Looks good, and it works!

I also came to the conclusion that selectors couldn't inherit from the same baseclass as Validators or Extractors, due to Selectors not being dependent on any "family" and, more importantly, not operating on Instances but instead operates on the Context and _creates_ instances.

@BigRoy If you have any way to work around this, I'd love to hear about it. In his pull-request, the validators and extractors are still inherited from Filter, where I also think they belong. But the Selector is not. This, in a way, breaks our concept of having them all interchangeable and sequential, as per our pipes and filters implementation, so we might have to have a look at how to go about it in some other way, or move away from that idea.

@tokejepsen Have you also tried this with a transform, as per your second selector? Also, we need to consider how to link selectors with their host; i.e. the current selectors are Maya-only so we will most likely have to append a `__hosts__` attribute to distinguish them.

mottosso:
> Maybe we should change that to look at which class (Validator/Extractor/Selector) it is coming from?

In what way did you find it limiting? This sounds reasonable, we could do that, but I'm also a fan of "if it quacks like a duck, it must be a duck" and not get too caught up in absolute inheritance members.

tokejepsen:
@mottosso, I did try the transforms selector with both your scene (https://github.com/abstractfactory/publish/tree/master/resources/tests/selection/multiple_selection_test/maya/scenes) and mine (https://github.com/abstractfactory/publish/tree/master/resources/tests/selection/tagging_test/sphere/library/characters/sphere). And they worked, when you of course, tag individual transforms with the indentifier. A `__hosts__` would be good, although I cant think off the top of my head how any plugin would able to be used in multiple hosts.

> In what way did you find it limiting? This sounds reasonable, we could do that, but I'm also a fan of "if it quacks like a duck, it must be a duck" and not get too caught up in absolute inheritance members.

The file structure and the reliance of prefixed strings seems weird and human error prone.

mottosso:
> A `__hosts__` would be good, although I cant think off the top of my how any plugin would able to be used in multiple hosts.

They most likely wouldn't. But imagine adding the path of plugins to available plugins, and Houdini finding a Maya plugin in there. It would need some way of knowing "Aha, this one isn't for me"

``` bash
plugins/
  select_maya_objectsets.py
  select_houdini_instances.py
```

mottosso:
> I did try the transforms selector with both your scene and mine

Ah, I see, you didn't save the scene with the attributes in place? I would recommend this, as it would allow anyone to open it up and study it. It's not as important currently, as there isn't that much to do, but down the line it will most likely become useful.

About the `select_transform.py`, this reminds me that we need a "debug" mode on the publish so we can see what's going on. But I noticed that the validators didn't trigger on bad naming convention. I.e. the inner names of that transforms aren't tagged with a three-letter extenstion - such as "_AST". Yet the validator didn't complain.

I took a look within the plugin and found that only the top-transform is added to the instance which means only it will get checked by the plugins. Maya of course exports everything within the group, so it looks like everything worked fine.

The same happens in the `select_object_set.py` plugin, and of course my original implementation.

Do you know if there is a way to only export selected objects, and disregard its hierarchy? That way, we would be damn sure it worked by looking at the result and would never get fooled this way. :)

tokejepsen:
@mottosso, sounds good to me. Made an issue to remind ourselves #36

mottosso:
Excellent, good initiative.

tokejepsen:
I'll save the scene next time:)

Cant think of any quick solution other than unparent/restore parent. 

One thing I'll need to test later is also a way to tell publish which selectors/validators/extractors to use, since I dont want to publish my model again in the rig file but it'll be tagged with the model familiy. Maybe this is something for the GUI to handle?

FYI, I tested the review extract and it did export a png sequence but they were empty. Something to fix for later #37

Issue #36, opened 2014-08-18
Implement __hosts__ attribute


tokejepsen:
Need a `__hosts__` attribute on plugins, to distingush between valid DCC hosts.

mottosso:
Having a look at this now.

mottosso:
About the prefix and suffix double underscores, I'm thinking of having these removed as they look plain ugly. Is anyone aware of any reason why these should exists? To me, the indicate a built-in method, such as `__init__` or `__str__`, as opposed to a user-added property.

``` python
# For example
class Filter(object):
    families = []
    hosts = []
    version = (0, 0, 0)
```

tokejepsen:
Got no problem with that:)

PR #38, opened 2014-08-18
Implementing #36

mottosso:
Also removed leading and following double underscores from `__family__` et. al., hence the many file-changes, along with implementing a preliminary `host()` function within domain.py to determine within which host plugins are currently being run from. Before, 'maya' was hard-coded within main.py.


tokejepsen:
works and looks good from my end.

PR #42, opened 2014-08-19
Maya isolated, review plugin fixed, and plugins moving home

mottosso:
Speaks for themselves.

### DIFF
--- a/publish/integration/maya/userSetup.py
+++ b/publish/integration/maya/userSetup.py
@@ -1,2 +1,40 @@
-from publish.main import eval_append_to_filemenu
+
+from maya import mel
+from maya import cmds
+
+
+def eval_append_to_filemenu():
+    """Add Publish to file-menu
+
+    .. note:: We're going a bit hacky here, probably due to my lack
+        of understanding for `evalDeferred` or `executeDeferred`,
+        so if you can think of a better solution, feel free to edit.
+
+    """
+
+    # As Maya builds its menus dynamically upon being accessed,
+    # we force its build here prior to adding our entry using it's
+    # native mel function call.
+    mel.eval("evalDeferred buildFileMenu")
+
+    script = """
+import publish.main
+
+cmds.menuItem('publishOpeningDivider',
+              divider=True,
+              insertAfter='saveAsOptions',
+              parent='mainFileMenu')
+cmds.menuItem('publishScene',
+              label='Publish',
+              insertAfter='publishOpeningDivider',
+              command=lambda _: publish.main.publish_all())
+cmds.menuItem('publishCloseDivider',
+              divider=True,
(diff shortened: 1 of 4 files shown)


mottosso:
Do have a look at [`userSetup.py`](https://github.com/mottosso/publish/blob/8da83230880f75322fd9b364ddbaccd895db9a8f/publish/integration/maya/userSetup.py) though, I had to run through some hacks to get the menu-item to show up properly, but I'm sure there is a better way.

BigRoy:
Just wanted to let you know that cmds.evalDeferred could be replaced with maya's executeDeferred from maya.utils

``` python
import maya.utils

def createMenu():
   pass

maya.utils.executeDeferred(createMenu)
```

That way you wouldn't need to wrap up mel commands. Should work just a nicely. :)

mottosso:
Thanks Roy, I did actually try that but couldn't manage to get the item to show still. I assumed it had to do with the order in which the two defers are executed, almost as if Python got executed before MEL, which would have caused Publish being added before the menu got built.

BigRoy:
Hmm, makes sense! :) I've heard of _issues_ with executeDeferred before, but we use it all over the place in userSetup.py at the studio. But this sounds like it could be one of those 'issues' that I read about on the internet. As long as it works we're getting somewhere, can always optimize/fix later.

mottosso:
Yeah, I would generally prefer it over evalDeferred too due to the benefit you spoke of above.

tokejepsen:
works on my end:) had to make the integration folder into a module to update my userSetup.py in my setup with:

``` python
from publish.integration.maya import userSetup
userSetup.eval_append_to_filemenu()
```

mottosso:
> works on my end:) had to make the integration folder into a module to update my userSetup.py in my setup

Ah, are you not having the `/integration/maya` folder on your PYTHONPATH? Are you instead adding that command to your `/Documents/maya/scripts/userSetup.py`?

That's something we should talk about too, in the [Installing](https://github.com/abstractfactory/publish/issues/31) issue, how to get from `pulling the repo` to `having Publish in your file-menu`. At the moment, I'm expecting users to append the `/integration/maya` folder to their PYTHONPATH, but that only applies to Maya. When Houdini gets an integration, it won't be making much use of `userSetup.py`

Another method would be to do what you've done here, to manually add it to your own `userSetup.py`. Ideally though I'd rather have it be fully automatic upon installing publish:

``` bash
$ pip install pypublish
$ # Done!
```

tokejepsen:
think I missed out on the PYTHONPATH method, will be using that now:)

mottosso:
Merging and documenting "domain.py" a little better in Wiki.
https://github.com/abstractfactory/publish/wiki/Quickstart#domainpy

PR #46, opened 2014-08-20
Selector fix and another Validator

tokejepsen:
Dunno why Git thinks the commits from @mottosso are different from the original repo.


mottosso:
If you have a look at the [Network](https://github.com/abstractfactory/publish/network) you can see that your latest pull by the time you submitted this pull-request is from an earlier version of the repo than the current one.

I'll do a merge on my end and update my request.

PR #52, opened 2014-08-23
Second pass on the configuration system

mottosso:
An updated configuration system, with cascading user-configuration.

In a nutshell, it mimics the configuration of Sublime and Maya amongst others in that it provides a default configuration, `config.json` and also looks for configuration in the users HOME directory. This allows users to store only modified variables in on their local computer that will override the behaviour or Publish, such as which identifier to look for within Maya (currently `publishable`) or which regex to use when looking for validators and so forth.

The configuration would persist inbetween updates of Publish and even across installs, and could be persisted either locally or on a central server so as to provide updates for an organisation; similar to `userSetup.py` of Maya.

From the doc-string or `config.py`:

``` python
"""Publish configuration

Configuration is copied into the users HOME directory
upon first use.

Usage:
    To alter the configuration of Publish to suit your needs,
    simply modify your own copy located at <HOME>/.publish

    It is a JSON-formatted, plain-text file.

    Members of the configuration is then accessed via dot-
    notation.

    >>> import publish.config
    >>> assert publish.config.identifier == 'publishable'

Attributes:
    identifier: How to distinguish between publishable instances
    prefix: Relative root-directory for extracted instances
    date_format: Format with which to date extracted instances
    paths_environment_variable: Which variable to look for added
        plugin paths.
    *_regex: Regular expression for finding the various plugins.

"""
```


mottosso:
Due to lack of comment-support of JSON, I'd like to propose moving to YAML for file-format, providing it as a vendor-package along with nose so as to not introduce any dependencies.

mottosso:
Sorry, that docstring was probably confusing, it came from an older version. This is more accurate:

``` python
"""Publish configuration

Usage:
    To alter the configuration of Publish to suit your needs,
    simply create your own copy at <HOME>/.publish as a JSON-
    formatted, plain-text file. E.g.

    # .publish
    {
        "identifier": "MyIdentifier"
    }

    Members of the configuration is then accessed via dot-
    notation.

    >>> import publish.config
    >>> assert publish.config.identifier == 'publishable'

Attributes:
    Attributes present within the config.json file:

    identifier:     How to distinguish between publishable instances
    prefix:         Relative root-directory for extracted instances
    date_format:    Format with which to date extracted instances
    paths_environment_variable: Which variable to look for added
        plugin paths.
    *_regex:        Regular expression for finding the various plugins.

"""
```

mottosso:
To make the behaviour crystal clear, here's a use-case:
1. User installs Publish
2. User creates `.publish` file at e.g. `c:\users\marcus\.publish`
3. User adds `{"identifier": "custom_publishable"}`
4. User runs Publish with custom identifier

Issue #49, opened 2014-08-21
Refactor Filter and Selector to both take Context as input


mottosso:
@tokejepsen mentioned [here](https://github.com/abstractfactory/publish/issues/40#issuecomment-52787305) in #40:

> Maybe we can even initialize both Validators and Selectors with a context, and get that same base class?

Which rang a bell in my head and I'll have a look at implementing this tomorrow morning.

Currently, Selectors are instantiated with a Context and adds Instances to it, whereas the other filters - validation, extraction and conform - are each instantiated with an Instance and operates on its members (the nodes in the scene).

In a way, we have been pre-determining which instances goes with which plugin. Instead, we could potentially pass the entire context to each validator and trust it to only pick instances relevant to itself. E.g. passing a context of 3 instances - e.g. 2 of family "model" and 1 of family "animation" - to a validator of family "model" would choose to only validate the 2 models, and leave the animation family alone.

tokejepsen:
>  Instead, we could potentially pass the entire context to each validator and trust it to only pick instances relevant to itself. 

We could also provide a function that filters the instances from a parent class. So in the validators we can call `self.instances`, and that will return the relevant instances.

mottosso:
> We could also provide a function that filters the instances from a parent class. So in the validators we can call self.instances, and that will return the relevant instances.

I'm giving this a go, but it looks like we'll have to refactor `abstract.py` for this to work.

For `Filter.instances` to work, `Filter` would need access to `domain.current_host()` which is not only in a separate module, a module that imports `abstract.py`, but is also something that reaches out onto the OS, breaking the encapsulation of the baseclasses.

The simplest thing I can think of would be to slap all modules together into a big `everything.py` where everything can have access to everything. But I'd really rather not.

Do you have any ideas?

tokejepsen:
I wasn't thinking the `self.instances` call through, so more of a wishful thinking I guess:)
I think it would be alright for now, to have the filtering of instances in each plugin. Anybody wanting to write a plugin will be looking through existing ones anyway to understand the process. Refactoring for the future if we find this to work out well.

mottosso:
No it was certainly a good idea, and I found it very necessary once I got into it.

The alternative is for the author of each plugin to handle filtering per family on his own, within each plugin, which produces a lot of duplicate code and would be easily prone to error. The only thing necessary should be the implementation of `process()` to handle a single Instance, as all other instances would be identical, similar to how it works at the moment.

[...]

PR #53, opened 2014-08-23
Attempt at #49

tokejepsen:
All plugins now inherit from the same Filter base class.
There is a convenience function for plugins, to get the relevant
instances called with `self.instances`.


mottosso:
Brilliant, this works. Some thoughts:

### Standalone function for instances()

What do you think about having Filter.instances() in the publish.plugin module as a standalone function? E.g.

``` python
def filter_by_instance(instance, plugins):
    """Return plugins matching an instance's criteria

    Arguments:
        instance (Instance): Instance with which to filter against
        plugins (list): List of plugins

    Returns:
        List of non-instantiated plugins.

    """

    plugins = list()

    for plugin in plugins:
        if instance.config.get('family') not in plugin.families:
            continue

        if instance.config.get('host') not in plugin.hosts:
            continue

        plugins.append(plugin)

    return plugins
```

Reasoning being that it could be used outside of Plugins too, like when we need the GUI to grab compatible plugins before actually running them.

Using it would then instead be:

``` python
plugins = publish.plugin.discover('validators')
compatible_plugins = publish.plugin.filter_by_instance(instance, plugins)
```

### process(context)

What do think about having `context` be part of the `process()` arguments, instead of the constructor for the class? E.g.

``` python
class Filter(object):
    def process(self, context):
        ..
```

That way, it would become more obvious what the implemented has to work with, as opposed to requiring knowledge about non-visible members of the class. Using it would then change:

``` python
# From
plugin(context).process()

# To
plugin().process(context)
```

tokejepsen:
> What do you think about having Filter.instances() in the publish.plugin module as a standalone function? 

Sounds good to have the code you posted outside certainly. But `Filter.instances()` looks like this;

``` python
def instances(self):
        result = []
        for instance in self.context:
            if instance.config.get('family') in self.families:
                result.append(instance)
        return result
```

So just to clarify this returns the relevant instances for the plugin, and not the other way around.

> What do think about having context be part of the process() arguments, instead of the constructor for the class?

I dont really have a problem with this, though you wouldn't be able to use the context outside of the process function in the plugin unless passed in. But I can't think of a reason why you want to do that anyway.

publish/abstract.py
```diff
@@ -58,21 +65,8 @@ class Selector(object):
...
-    @abc.abstractproperty
-    def version(self):
-        return tuple()
-
-    def __init__(self, context):
-        self.context = context
-
-    @abc.abstractmethod
-    def process(self):
+    @property
```
mottosso: This doesn't need to be here as it is inherited from Filter.

mottosso: Sorry, haven't used this feature before, I meant line 69, where you declare `def families(self)`

mottosso:
> So just to clarify this returns the relevant instances for the plugin, and not the other way around.

Ah! Good point, I thought of it the other way. You are right, I think we need it to be this way.

> I dont really have a problem with this, though you wouldn't be able to use the context outside of the process function in the plugin unless passed in.

Exactly, I think this is a fair trade I think as it makes it easier to understand how to fill in `process()`. It is always possible to pass the context around, like you said.

tokejepsen:
> This doesn't need to be here as it is inherited from Filter.

Does this mean I don't need to modify my code? Its done in the pull request here?

publish/abstract.py
```diff
@@ -38,16 +38,23 @@ def __str__(self):
...
-        self.instance = instance
+    def __init__(self, context):
+        self.context = context
         self.errors = list()
 
     @abc.abstractmethod
     def process(self):
         return None
 
+    def instances(self):
```
mottosso, line 45: What do you think about making this a `@property`? My motivation lies in sticking with verbs for functions/methods, and nouns for attributes/properties. That is, either going with "compute_instances()" or something similar, or:

``` python
@property
def instances(self):
    ...
```

mottosso:
> Does this mean I don't need to modify my code? Its done in the pull request here?

I think these are just comments. We probably can't affect the code via GitHub this way.

tokejepsen:
> What do you think about making this a @property?

I'll try this. Completely removed it from any decorators to make it work, but a `@property` should work.

> I think these are just comments. We probably can't affect the code via GitHub this way.

I'll make those change now.

mottosso:
> I'll make those change now.

Cool, these are just ideas though, throwing them out there mostly for discussion. But if you like'm, so do I.

tokejepsen:
> Sorry, haven't used this feature before, I meant line 69, where you declare def families(self)

Actually I was having problems with this. I can't seem to instantiate the Selector plugins, if they don't have the `self.families` defined. And that's why I did it in the parent class, so it wasn't needed in the child classes.

mottosso:
Ah, good point. Families doesn't apply to selectors. Looks like we've got a gap in our logic here..

tokejepsen:
> Ah, good point. Families doesn't apply to selectors. Looks like we've got a gap in our logic here..

Could just have the families property in the Validator and Extractor subclasses?

mottosso:
We'll probably have to cut families from Filter, and add it to Validator, Extractor and Conform individually. Thoughts?

tokejepsen:
> We'll probably have to cut families from Filter, and add it to Validator, Extractor and Conform individually. Thoughts?

I agree:)

mottosso:
Great minds think alike!

tokejepsen:
> We'll probably have to cut families from Filter, and add it to Validator, Extractor and Conform individually. Thoughts?

I'm not entirely sure about meta classes, so would the families property be decoractered like; `@abc.abstractproperty` or just `@property`?

mottosso:
> I'm not entirely sure about meta classes, so would the families property be decoractered like; @abc.abstractproperty or just @property?

Good question. Base-classes should use the abstractproperty, so as to guarantee that subclasses implement it. In this case, we want subclasses to implement this, so the baseclass should use abstractproperty. Each plugin however should be using the regular `@property`.

The idea is for Python to throw an error when a subclass is incomplete. That's what the abc.abstractproperty is for.

As a side-note, if you add the word #49 in your commit, GitHub will put a note in that issue. If you put `fixes #49`, then it will automatically close the issue too.

It's a useful way of keeping issues connected with commits. :)

tokejepsen:
> As a side-note, if you add the word #49 in your commit, GitHub will put a note in that issue.

Guess it doesn't do that when its in the title?

Updated the pull request, have another look.

mottosso:
> Guess it doesn't do that when its in the title?

No, you're right. I hadn't updated my browser. Sorry about that.

> Updated the pull request, have another look.

Looks good. Publishing works still too? Anything else you'd like to add before I merge?

tokejepsen:
> Looks good. Publishing works still too? Anything else you'd like to add before I merge?

Publishing still works. Think that is it for me for now.

mottosso:
(Not sure where merge comments go)

Merged, good work!

PR #65, opened 2014-09-01
Convenience method for compatible instances.

tokejepsen:


mottosso:
Toke is on fire! :)

This method got separated into plugin.py as a standalone function - `instances_by_plugin` - as it was needed elsewhere (in main.py) and for symmetry with its opposite - `plugins_from_instance`

https://github.com/abstractfactory/pyblish/blob/master/pyblish/backend/plugin.py#L283

Do you think you could use that instead?

tokejepsen:
Sure:) i'll have a look at adjusting the plugins.

tokejepsen:
closing this, as I'm coming back after major updates

Issue #59, opened 2014-08-28
Extraction plugins are solely responsible for placing files


mottosso:
This is very powerful, but won't do much good if each plugin stores data in different locations.

What would be better, is if the baseclass for Extraction could provide a mechanism with which to finally place file on disk where they belong, so that a Conform plugin could take it from there.

``` python
# Currently
class MyExtractor(Extractor):
     def process(self, context):
         # Produce file in a temp directory
         files = ['file1', 'file2']
         # Move files to extraction destination
         shutil.copy('/my/unique/directory')

# Ideally
class MyExtractor(Extractor):
     def process(self, context):
         # Produce file in a temp directory
         files = ['file1', 'file2']
         # Move files to extraction destination
         self.commit(files)
```

tokejepsen:
I was wondering about this the other day, as I wanted to look into making a Conformer. Won't the context need this information, for the Conformer to pick up down the line?

mottosso:
For plugins to accurately place files relative the working file, it must somehow have access to where the working file is located.

The end result is documented here:
- https://github.com/abstractfactory/pyblish/wiki/Extraction

One method could be to pass this information into the context when it is first being created.

``` python
def publish_all():
    current_path = cmds.workspace(dir=True, query=True)
    context  = pyblish.backend.plugin.Context(current_path=current_path)
```

A plugin could then use the information.

``` python
class MyExtractor(Extractor):
     def process(self, context):
         final_path = os.path.join(context.current_path, 'published')
```

It does however mean a dependency between `publish_all()` and Maya, in which case we'll need a Maya-specific implementation of it.

``` python
pyblish_maya.main.publish_all()
```

> I was wondering about this the other day, as I wanted to look into making a Conformer. Won't the context need this information, for the Conformer to pick up down the line?

Yes, it most certainly does. We'll have to find a good way to distribute this information, without getting tangled.

mottosso:
It struck me that this isn't an isolated requirement. Conform will also need a bunch of information, e.g. the source-file from which an instance was published, for reference to other users.

So, how about making the "additional data" gathering part of the Selection process? A selector could gather information about the scene, such as the workspace and file-name and such.

A pair of default selectors could gather the global information such as time of publish, currently logged on user and other data.

In the end, the context would have access to this information and if at any point in time a user makes more plugins that depends on further information, they could just make another selector to gather it up.

[...]

Issue #39, opened 2014-08-18
Plugins Logging


tokejepsen:
Fiddling with a validator now, wanted some feedback from the plugin so I just used a print statement. Shouldn't we implement a logging from the plugins?

mottosso:
At the moment, there is a logging class-decorator which adds a `self.log` to affected classes. See [here](https://github.com/abstractfactory/publish/blob/master/publish/plugins/extract_model.py#L12)

``` python
# Example
import publish.lib

@publish.lib.log
class MyClass(...):
    def process(self):
        self.log.info("Hello world!")
```

At the moment, the logging isn't being initialised in Publish, I've relied on Maya to do this so far. I've had issues with logging and Maya in that when Publish initialises one then Maya logs twice. Do you know any way around that? 

Here is how I usually setup logging.

``` python
def get_formatter():
    formatter = logging.Formatter(
        '%(asctime)s - '
        '%(levelname)s - '
        '%(name)s - '
        '%(message)s',
        '%H:%M:%S')
    return formatter


def setup_log():
    log = logging.getLogger('publish')

    if log.handlers:
        return log.handlers[0]

    log.setLevel(logging.INFO)

    formatter = get_formatter()
    stream_handler = logging.StreamHandler()
    stream_handler.setFormatter(formatter)
    log.addHandler(stream_handler)

    return log
```

tokejepsen:
I have little to no experience with logging. Really my first project with logging:)

I'll investigate a bit more.

mottosso:
Ok, no worries. :) Just append the decorator to your validator, and use the `self.log` member. We can tailor logging afterwards, but this can remain the interface until we find something better.

mottosso:
In the current version, there is some initialisation of logging happening in [`__init__.py`](https://github.com/abstractfactory/publish/blob/143c80010e69c78230658aec79460f4bdd6d1579/publish/__init__.py). If you notice in Maya, each logging entry happens twice because of this. If logging isn't initialised, then it won't produce any printing outside of Maya, and possibly not in other hosts either as I believe Maya is being particularly "smart" in this regard.

If anyone can think of a reason for this or how to solve it, feel free to have a look.

mottosso:
This has been fixed in 1.0.7!

PR #66, opened 2014-09-02
Implementing commit()

### LINKED ISSUE #58: Provide debug prints for plugins
At the moment, it's quite tricky to see what's going on when running the various plugins.

Each plugin is responsible for providing its own debug information, which is cool, but doesn't provide consistency and can easily be forgotten. Instead, we could provide built-in statements - like "Running.." and "Passed.." - that get run automatically, alongside any custom messages produced by each individual plugin

``` python
# The super() method could hook into already-implemented logging
class MyValidator(Validator):
    def process(self, context):
        super(MyValidator, self).process(context)

# Or, the baseclass could provide manual hooks
class MyValidator(Validator):
    def process(self, context):
        self.log_start()
        # ...
        self.log_finished()

# Or, a metaclass could append functionality implicitly
class MyValidator(Validator):
    def process(self, context):
           # ...
```

### LINKED ISSUE #62: config.json has old brandname
``` json
{"paths_environment_variable": "PUBLISHPLUGINPATH"}
```

It should say "PYBLISHPLUGINPATH", note the **Y**
- [Link](https://github.com/abstractfactory/pyblish/blob/1166405cd08863165832072f677564f1a3876642/pyblish/backend/config.json#L6)

mottosso:
Mainly addressing #59, but touching on other issues along the way.
- Fixed #58
- Fixed #62
- Addressed #39

See commit comments for full change log.


mottosso:
One of the things I've also added is a `create_instance` on the Context object.

The idea is that Instances should have access to their parent Context and have as such been given a `context=None` argument in their constructor.

``` python
class Instance(...):
    def __init__(self, name, context=None):
        self.name = name
        self.context = context
```

However, as a context is usually available when creating instances, `create_instance` eliminates some typing. E.g.

``` python
# From this
>>> ctx = pyblish.backend.plugin.Context()
>>> inst = pyblish.backend.plugin.Instance(name='my_name', context=ctx)
>>> ctx.add(inst)

# To this
>>> ctx = pyblish.backend.plugin.Context()
>>> inst = ctx.create_instance(name='my_name')
```

Accessing its parent context object is done [here](https://github.com/mottosso/pyblish/blob/e44a1fceeb703ad0088c1529403d81b1489e2f6f/pyblish/backend/plugin.py#L153), I could see no other way.

mottosso:
Merging and moving on!

Issue #67, opened 2014-09-05
Plugins within identical file-names are duplicated


mottosso:
When importing plugins from a module `select_environment.py` located within one directory and another file of identical name within another directory the contents of said module is appended to rather than replaced.

I've narrowed it down to a caveat within the `imp` module which is used for dynamic loading of Python modules.

``` python
>>> mod1 = imp.load_source(name='select_environment', 
                           pathname='c:\some\path\select_environment.py')
>>> id(mod1)
55555551
>>> dir(mod1)
UniqueClass1
UniqueClass2
>>> mod2 = imp.load_source(name='select_something_else', 
                           pathname='c:\some\path\select_something_else.py')
>>> id(mod2)
55555552
>>> dir(mod2)
UniqueClass3
```

This is what we would expect, the two modules take up separate portions of memory and are thus completely separate objects. Loading one does not have any affect on the other; i.e. there are no side-effects.

``` python
>>> mod1 = imp.load_source(name='select_environment', 
                           pathname='c:\some\path\select_environment.py')
>>> id(mod1)
55555551
>>> mod2 = imp.load_source(name='select_environment', 
                           pathname='c:\some\other\path\select_environment.py')
>>> id(mod2)
55555551
```

> Note the difference in its `pathname`, but identical `name`.

Here on the other hand, memory locations are identical, suggesting that `imp` has replaced `mod1` with `mod2`. What really happens however is a merge.

``` python
>>> mod1 = imp.load_source(name='select_environment', 
                           pathname='c:\some\path\select_environment.py')
>>> dir(mod1)
UniqueClass1
UniqueClass2
>>> mod2 = imp.load_source(name='select_environment', 
                           pathname='c:\some\other\path\select_environment.py')
>>> dir(mod2)
UniqueClass1
UniqueClass2
UniqueClass3
```

> Note that `UniqueClass1` and `UniqueClass2` is present in both `mod1` and `mod2`

This caused plugins to accumulate and cause lots of duplicated plugins to be processed whenever they were sharing names.

tokejepsen:
Wow, thats quite a bit of detective work there:) Have you got a fix in mind?

mottosso:
Heh, yeah thought I was going mad. :) I'm working on it, but at the moment it seems the simplest way to go is to ensure unique names of modules, which seems to be working fine so far.

The reason we haven't caught it until now is because we've never checked for duplicate plugins before. We haven't had that many so it hasn't been noticeable.

[...]

PR #68, opened 2014-09-05
Issue 67

mottosso:
I've refrained from using `imp` and instead rely on good old fashioned sys.path to do the work for us. It's a little brute, but there are no more surprises. The merging mentioned in #67 is simply over my head and I can't make either heads or tails of it.

Funnily enough, as a result of hindering duplicated modules, one of the Maya plugins stopped working as it was also duplicated. The `Deduplicating plugins` commit handles that little caveat. :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-- 2.6
 - 2.7
 install:
 - pip install coveralls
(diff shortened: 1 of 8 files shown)


mottosso:
We've also been able to support Python 2.6 in addition to 2.7, which means Maya 2011 or so and below. I had to disable this due to how the import mechanism works currently, but we'll get it going again once things get more stable.

mottosso:
Merging this as it is mainly a bugfix.

PR #69, opened 2014-09-06
Improved plugin interface

mottosso:
A much simplified interface for plugins with methods separated for Instance and Context arguments.
https://groups.google.com/forum/#!topic/pyblish/tcY3SBK-NHQ

Plugins that implement `process()` will still work as-is, the method has simply gotten a base class implementation that triggers `process_instance()` and `process_context()`.

``` python
# Psuedo-code
def process(self, context):
    self.process_context(context)
    for instance in context:
        self.process_instance(instance)
```

In a nutshell, plugins that modify the context only need to override `process_context()` and plugins that modify the instance override `process_instance()`. Some plugins modify both, and that is okay.

`process_instance()` comes with the added benefit of getting only compatible instances as argument.

``` python
class ValidateModelFamily(Validator):
    families = ['model']
    def process_instance(self, instance):
        # instance is of family "model"
```
# Data

Additionally, collected data is now fully part of the Selection process, which works really well! What this means is that, rather than validators or extractors gathering information from the scene or system, selectors gather this information and stores it within each instance or context. The information includes:
- System time and date
- Currently logged on user
- Current working directory
- Current working file
- Optional workspace

The data may then be used by plugins and they are guaranteed to get the same data, which may vary with e.g. collecting `date` if one plugin executes a few seconds after the first. 

It also means we can start making more defined assumptions and requirements from and in our plugins. For example, a conform plugin that takes data from a temporary location as provided by an extractor will need to know where to find this information. This allows us to set up rules for how plugins communicate with each other. Extractor plugins can now communicate with Conform plugins via members of data.

``` python
# Extractor
instance.set_data('temporary_location', value='c:\temp\data_here.mb')

# Conform
shutil.copytree(instance.data('temporary_location'), '\\server\final_destination')
```

To get started, I included a few data selectors.
https://github.com/mottosso/pyblish/tree/master/pyblish/backend/plugins

Best,
Marcus

### DIFF
--- a/pyblish/backend/config.json
+++ b/pyblish/backend/config.json
@@ -8,5 +8,7 @@
     "validators_regex": "^validate_.*\\.py$",
 	"extractors_regex": "^extract_.*\\.py$",
 	"selectors_regex": "^select_.*\\.py$",
-	"conforms_regex": "^conform_.*\\.py$"
+	"conforms_regex": "^conform_.*\\.py$",
+
+    "commit_template": "{prefix}/{date}/{family}/{instance}"
 }
--- a/pyblish/backend/plugin.py
+++ b/pyblish/backend/plugin.py
@@ -58,19 +58,14 @@ registered_modules = list()
 log = logging.getLogger('pyblish.backend.plugin')
 
 
+@pyblish.backend.lib.log
 class Plugin(object):
-    """Abstract base-class for plugins
-
-    Attributes:
-        hosts: Hosts compatible with plugin
-        version: Current version of plugin
-
-    """
+    """Abstract base-class for plugins"""
 
     __metaclass__ = abc.ABCMeta
 
-    hosts = list()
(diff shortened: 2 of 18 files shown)


mottosso:
Do also have a look at the improved [`commit()`](https://github.com/mottosso/pyblish/blob/5b2aeb85f05ca6fc5c062f098dcfb42d07cb46f3/pyblish/backend/plugin.py#L183) from issue #59. It illustrates the communication taking place between extractor and Plugin baseclass. In a nutshell, the extractor exposes data that is expected within the baseclass. If the extractor doesn't expose this data, e.g. in a faulty user-implementation, it simply cannot perform the commit.

Committing in this case relates to storing data relative the current working file.

``` python
# From here
c:\user\marcus\temp\cdfakfxasl\publish_SEL

# To here
//projects/spiderman/Peter/marcus/maya/published/publish_SEL
```

PR #70, opened 2014-09-06
Yaml Config

mottosso:
Upgrading configuration to YAML due to better support for comments.

> The large amount of file-changes is simply due to including the YAML package. The change on our end is 2 files, `config.py` and `config.json` -> `config.yaml`

In short, it went from this:

**config.json**

``` json
{
    "prefix": "published",
    "identifier": "publishable",
}
```

To this:

**config.yaml**

``` yaml
# The identifier determines the predicate with which
# to distinguish between content within a host that 
# is publishable from that which is not.
identifier: publishable

# The prefix determines within which subdirectory
# published instances are to be stored. This is
# ultimately used within the `commit_dir` path
# template below.
prefix: published
```

Nothing else has changed and the interface is the same.

``` python
>>> from pyblish.backend import config
>>> config.identifier
'publishable'
```


mottosso:
Travis is failing due to getting confused and looking at the wrong branch.

PR #71, opened 2014-09-06
Coverage py

mottosso:
This request appends the ability to run coverage tests on your local machine.

If you double-click the `run_coverage.bat`, it will generate a directory with a web interface in which you can spot which lines of code are covered and which are not, along with the current percentage of code covered.

This is useful when trying to reach 100% coverage. Generally, whenever you add or remove functionality, we would ideally try to maintain or improve upon the coverage.

More information can be found here:
http://nedbatchelder.com/code/coverage/

> The directory won't be included in commit, which is a good thing as it's generated each time you run the command.


mottosso:
Merging in preparation for documentation and tutorials.

Issue #110, opened 2014-09-28
Plugin execution doesn't respect hosts property


tokejepsen:
Executing a plugin in Nuke with `hosts = ['maya']` proceeds to process

mottosso:
Solved in #111

Issue #95, opened 2014-09-17
Repair


mottosso:
### Goal

Implement counter-measures in your validators to automatically repair any errors found.
### Architecture

Upon validator failure, present the user with the option to repair. The counter-measure is implemented within the validator itself.

``` python
class ValidateNoMutedChannels(Validator):
    def repair_instance(self, instance):
           """Muted channels were found"""
           # Delete mute-nodes
```

tokejepsen:
I was wondering whether we could use a temporary solution, have a "Repair Instances" or "Repair Validations" button in the hosts under the "Validate".
This will just run the `repair_instance` method on the Validators.

mottosso:
Implementing this next. How about something like this?

``` python
import pyblish.main
pyblish.main.publish_all(context, auto_repair=True)
pyblish.main.validate_all(context, auto_repair=True)
```

I'd rather not create more menu-items, but rather leave the more advanced options to the GUI.

tokejepsen:
So will the auto_repair will be on by default?

mottosso:
As a user, I wouldn't expect a publish to affect my scene unless I explicitly told it to, so I would keep it optional. What do you think?

ghost:
+1

## 

David Martinez - Technical Animator

Email: david.martinez.anim@gmail.com
Website: http://www.elusiveideas.com

tokejepsen:
I agree, but I was trying to find a solution for getting to use the repair
feature without having to wait for the gui to be finished?

mottosso:
You can customise your menu-item. Just use the code above in-place of the existing code.

mottosso:
Okay guys, this has been implemented. If you encounter any issues, let's re-open this issue.

Repair messages are logged along with their respective plug-in.

``` bash
ValidateAssetDefaults (Validate Asset Defaults)...
#     ERROR    Some nodes had values not returned to their defaults: 
#     ERROR        R_arm_CTL.[u'rotateX', u'rotateY', u'rotateZ']
#     WARNING  There were errors, attempting to auto-repair..
#     INFO     Auto-repair successful
```

If auto-repair succeeds, validation is considered successful and the publish resumes. Otherwise, it fails like when validation fails, with messages from both validation and the repair.

PR #111, opened 2014-10-01
Implemented #95

mottosso:
### Auto-repair

You can now specify `auto_repair=True` in pyblish.main.publish() and pyblish.main.validate_all() and plug-ins with a `repair_instances` function will process instances that fail validation.


mottosso:
This also fixes #110

PR #122, opened 2014-10-11
1.0.9 - Requires

mottosso:
Hi all,

Small update for the future - Requires.
### Requires

`requires` is an attribute of a plug-in specifying with which version of the Pyblish API the plug-in is compatible with. This is so that if you have plug-ins that pre-date a particular version since a breaking change - such as v2.0 - the discover mechanism will recognise this and log a message about what has happened.

This way, we can keep a mixture of plug-ins available, without any of them causing any trouble. The idea is for you to be able to update Pyblish, whilst later on update your plug-ins, should you want to.

It looks like this.

``` python
class ValidatePullRequest(Validator):
    requires = "pyblish>=1.0.8"
```

This plug-in requires a version of Pyblish _above or equal_ to 1.0.8. Essentially saying, if you plug-in is discovered by 1.0.9, it won't get loaded.

The syntax is coming from the standard PyPI requirements.txt, and the library responsible for making comparisons, along with its documentation, can be found here.
- https://github.com/mottosso/iscompatible
- http://iscompatible.readthedocs.org/en/latest/
### Later on

The idea is then to allow for multiple criterias.

``` python
class ValidatePullRequest(Validator):
    requires = ["pyblish>=1.0.8", "pyblish-maya>=0.9, <=1.1.2"]
```

This one requires both Pyblish up to 1.0.8 and Pyblish for Maya, versioned between 0.9 and 1.1.2. This is so that we can design plug-ins with intricate requirements that say "No, I need extension X in order to work", allowing users to re-use as much as possible of other extensions, whilst still keeping their hair.
- See [CHANGES](https://github.com/mottosso/pyblish/blob/d2050b7c2e3e67037c72ec1f872b11bf76c45992/CHANGES) for more.

Best,
Marcus

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,6 @@ dist
 *.pyc
 cover
 .coverage
-docs/_**
+docs/_*
 build
 playground.py
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ cover
 docs/_*
 build
 playground.py
+*.sublime-*
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,19 @@ pyblish Changelog
 
 This contains all major version changes between pyblish releases.
 
+Version 1.0.9
+-------------
+
+- Requires. Plug-ins may now specify a version with which they
+    are compatible with, using [iscompatible][] which is a
+    requirements.txt-like syntax for dependency specifications.
+- Improved logging, including visualisation of which instance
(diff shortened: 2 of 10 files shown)


mottosso:
More info in the usergroup.
- https://groups.google.com/forum/#!topic/pyblish/qCgYvlIi3WE

PR #166, opened 2015-03-19
1.0.13

mottosso:
See CHANGES


mottosso:
`util.py` now represents user-facing convenience functions, such as `publish()` and `select()`. Having them in a file called `main.py` gave the wrong idea that it was somehow tied to core functionality, when in reality these functions are merely convenience functions that wrap certain parts of the core functionality, primarily that found in `plugin.py` and `api.py`.

PR #180, opened 2015-05-20
rearranging the check for current host.  to work with Houdini

larsbijl:
if houdini has the same version of python as the machine (centos 7 for example) it will default to the base python. so checking to see if hou is imported can tell us.

also moving python down because I think it should be from the most specific to the most broad.

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -665,10 +665,6 @@ def current_host():
 
     executable = os.path.basename(sys.executable).lower()
 
-    if "python" in executable:
-        # Running from standalone Python
-        return "python"
-
     if "maya" in executable:
         # Maya is distinguished by looking at the currently running
         # executable of the Python interpreter. It will be something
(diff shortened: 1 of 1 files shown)


mottosso:
That's a good idea. We've been having problems with getting Houdini recognised as well, primarily when running it from any other executable than `houdini` or `houdinifx`, such as `hescape`.

This should solve that.

I also had a thought, at the moment, if all else fails, and `python` _isn't_ in the name, then an exception is thrown. What do you think about having that as a true catch-all, and simply return `python` when none of the above tests work?

We're also looking into an improved method of identifying hosts here:
- http://forums.pyblish.com/t/hosts-integration/75

larsbijl:
I think it's fine to assume that python will be the fallback.  seeing as the integration is already running in Python just saying it's a vanilla would not cause a major problem.  it might just lose some functionality when publishing.  

however having this is an extra option so that you can turn it on in case you want to be very strict on how things get published might make sense in the larger facility.

mottosso:
> however having this is an extra option so that you can turn it on in case you want to be very strict on how things get published might make sense in the larger facility.

That's a good point. I can imagine it being confusing, developing a plug-in for `softwarex` but not seeing it show up because the host was identified as being `python`.

As it's currently strict, I think we should leave it in, and potentially open it up further down the line. It's easier to open up than to plug.

Merging.

Issue #231, opened 2015-11-12
Instance multi-family support


mottosso:
![image](https://cloud.githubusercontent.com/assets/2152766/11120293/9c8a160c-8946-11e5-9eb6-5802ee1b03fa.png)
### Goal

Facilitate the use of external plug-in(s) in existing publishing pipeline.
### Motivation

Installing someone else's plug-ins currently requires you to architect your content according to the families provided by those plug-ins.

That's fine, assuming this is the first and only series of plug-in you install and that you start from scratch.

But what if you already have a series of plug-ins and things are working well? You are just interested in appending a few from an external source that help with one or more issues?

Well, with this you can.
### Implementation

An implementation is fairly straightforward, an Instance is already capable of hosting arbitrary data so we'll add to that.

``` python
instance.data["families"] = ["geometry", "human", "primary"]
```

The mechanism currently in charge of determining compatibility between plug-in and instance is called [instances_by_plugin](https://github.com/pyblish/pyblish/blob/14ea713546c5431b294b9cda0d210699b5c1b5b1/pyblish/logic.py#L267) and will be augmented to also consider the new `families` data entry.

For backwards compatibility, the current `family` will remain and involves no overhead and little cost to maintenance (about 4 lines of code at a near location).

> Thanks to @BigRoy for the suggestion.

tokejepsen:
Would you not need to have the collectors the same as well, so the data is the same in the instance?

mottosso:
The collectors would be the ones assigning multiple families, this is primarily for subsequent plug-ins, where you are still in control of what the collector does. Like validators and extractors, possibly integrators too.

Here's how I see this work.
1. Install new validator, with support for `familyX`
2. Append `familyX` to relevant instances in your Collector.

Or better yet, have your collector remain general enough to derive `familyX` from the content somehow, such as picking it up from a user-defined attribute in Maya; as in Magenta.

mottosso:
Implemented in [1.2.2](https://github.com/pyblish/pyblish/releases/tag/1.2.2).

PR #232, opened 2015-11-12
Implemented #231

mottosso:
Adding support for instances with multiple-families (#231)


mottosso:
I've written this primarily for the developer docs, it hasn't yet been tested with QML or anything and might not work very well.

mottosso:
This has been implemented in 1.2.2, [here](https://github.com/pyblish/pyblish/pull/236).

PR #235, opened 2015-12-03
Fix time() returning local time but claiming to return UTC.

jedfrechette:
The time function was returning a string containing the local time with the UTC timezone designation 'Z' incorrectly appended. I've changed this to return the current UTC time. Since the manual strftime call was producing the same output as the builtin isoformat function I also opted to use the builtin function.

### DIFF
--- a/pyblish/lib.py
+++ b/pyblish/lib.py
@@ -65,7 +65,8 @@ def extract_traceback(exception):
 
 
 def time():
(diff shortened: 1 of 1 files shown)


mottosso:
Well spotted!

Referencing original issue for completeness.
- https://github.com/pyblish/pyblish/issues/181

Thanks @jedfrechette

PR #238, opened 2016-01-14
Data Handlers

tokejepsen:
This introduces data handlers, which handle data coming from pybilsh-qml.

The test for this is in pyblish-rpc, so when running [pyblish-qml/app.py](https://github.com/pyblish/pyblish-qml/blob/master/pyblish_qml/app.py) you see "Toggle Instance" and "ValidateTogglePlugin". When toggling the instance and/or the plugin, a message should get printed to the console similar to:

```
Changed Toggle Instance to False
```

The data handler is in [pyblish-rpc/mocking.py](https://github.com/tokejepsen/pyblish-rpc/blob/a62ae6601d64b1d70de321b9e0de2f04a27b0b62/pyblish_rpc/mocking.py#L513-L519), which triggers functions on the [instance](https://github.com/tokejepsen/pyblish-rpc/blob/a62ae6601d64b1d70de321b9e0de2f04a27b0b62/pyblish_rpc/mocking.py#L92-L96) and the [plugin](https://github.com/tokejepsen/pyblish-rpc/blob/a62ae6601d64b1d70de321b9e0de2f04a27b0b62/pyblish_rpc/mocking.py#L105-L108).
Each data handler runs when data is send from pyblish-qml in [pyblish-rpc/service.py](https://github.com/tokejepsen/pyblish-rpc/blob/a62ae6601d64b1d70de321b9e0de2f04a27b0b62/pyblish_rpc/service.py#L127-L143). Here I'm converting data from pyblish-qml to classes. The arguments for a data handler is:
- **item**
  - Either instance or plugin
- **data_item**
  - The name of the item in the pyblish-qml model, meaning the attribute. In the example I'm specifying which attribute to act upon in the data handler; https://github.com/tokejepsen/pyblish-rpc/blob/a62ae6601d64b1d70de321b9e0de2f04a27b0b62/pyblish_rpc/mocking.py#L515
- **new_value**
  - The value the attribute has changed to.
- **old_value**
  - The value the attribute was changed from.

As it stands pyblish-qml only send data across on toggling an item; https://github.com/tokejepsen/pyblish-qml/blob/b6f11235fc3d8450af4961dc83f4cd27a827cbcf/pyblish_qml/control.py#L457-L458

Let me know if anything is unclear:)

Currently I don't like that you need to use a static method on a plugin, but we can probably figure out a different workflow.
Overall I think its a bit over complicated, but it does lay down the foundation for saving any data to the scene.

### DIFF
--- a/pyblish/lib.py
+++ b/pyblish/lib.py
@@ -363,15 +363,14 @@ def emit(signal, **kwargs):
             signal (string): Name of signal emitted
 
         Example:
-            def myCallback(data=None):
-                print data
-
-            pyblish.api.register_callback("someSignal", myCallback)
-            emit("someSignal", data={"something": "cool"})
+            >>> import sys
+            >>> from .plugin import register_callback
+            >>> register_callback("mysignal", lambda data: sys.stdout.write(data))
+            >>> emit("mysignal", data={"something": "cool"})
+            {'something': 'cool'}
     """
     for callback in pyblish._registered_callbacks.get(signal, []):
         try:
             callback(**kwargs)
         except Exception as e:
             traceback.print_exc(e)
-            raise Exception(e)
--- a/pyblish/lib.py
+++ b/pyblish/lib.py
@@ -5,7 +5,7 @@ import logging
 import datetime
 import traceback
 
-import pyblish
+from . import _registered_callbacks
 
 _filename_ascii_strip_re = re.compile(r'[^-\w.]')
 _windows_device_files = ('CON', 'AUX', 'COM1', 'COM2', 'COM3', 'COM4',
@@ -354,22 +354,23 @@ def import_module(name, package=None):
 
 
 def emit(signal, **kwargs):
-    """Searches through the registered callbacks and executes any callback that
-    matches the signal.
-    The keyword arguments (kwargs) are passed directly through from the emitter
-    to the callbacks.
+    """Trigger registered callbacks
 
-        Arguments:
-            signal (string): Name of signal emitted
+    Keyword arguments are passed from caller to callee.
 
-        Example:
-            >>> import sys
-            >>> from .plugin import register_callback
-            >>> register_callback("mysignal", lambda data: sys.stdout.write(data))
-            >>> emit("mysignal", data={"something": "cool"})
-            {'something': 'cool'}
+    Arguments:
+        signal (string): Name of signal emitted
+
+    Example:
+        >>> import sys
+        >>> from .plugin import register_callback
+        >>> register_callback("mysignal", lambda data: sys.stdout.write(data))
+        >>> emit("mysignal", data={"something": "cool"})
+        {'something': 'cool'}
+
     """
-    for callback in pyblish._registered_callbacks.get(signal, []):
+
+    for callback in _registered_callbacks.get(signal, []):
         try:
             callback(**kwargs)
         except Exception as e:
--- a/pyblish/lib.py
+++ b/pyblish/lib.py
@@ -5,6 +5,8 @@ import logging
 import datetime
 import traceback
 
+import pyblish
+
 _filename_ascii_strip_re = re.compile(r'[^-\w.]')
 _windows_device_files = ('CON', 'AUX', 'COM1', 'COM2', 'COM3', 'COM4',
                          'LPT1', 'LPT2', 'LPT3', 'PRN', 'NUL')
@@ -349,3 +351,27 @@ def import_module(name, package=None):
     __import__(name)
 
     return sys.modules[name]
+
+
+def emit(signal, **kwargs):
+    """Searches through the registered callbacks and executes any callback that
+    matches the signal.
+    The keyword arguments (kwargs) are passed directly through from the emitter
+    to the callbacks.
+
+        Arguments:
+            signal (string): Name of signal emitted
+
+        Example:
+            def myCallback(data=None):
+                print data
+
+            pyblish.api.register_callback("someSignal", myCallback)
+            emit("someSignal", data={"something": "cool"})
+    """
+    for callback in pyblish._registered_callbacks.get(signal, []):
+        try:
+            callback(**kwargs)
+        except Exception as e:
+            traceback.print_exc(e)
+            raise Exception(e)
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -482,8 +482,8 @@ def process(plugin, context, instance=None, action=None):
             provider.invoke(runner)
             result["success"] = True
     except Exception as error:
-        emit("pluginFailed", plugin=plugin, context=context, instance=instance,
-             error=error)
+        pyblish.lib.emit("pluginFailed", plugin=plugin, context=context,
+                         instance=instance, error=error)
         pyblish.lib.extract_traceback(error)
         result["error"] = error
 
@@ -783,6 +783,12 @@ def current_host():
 
 
 def register_callback(signal, callback):
+    """Register a new callback
+
+    Arguments:
+        signal (string): Name of signal to register the callback with.
+        callback (func): Function to execute when a signal is emitted.
+    """
 
     if signal in pyblish._registered_callbacks:
         pyblish._registered_callbacks[signal].append(callback)
@@ -790,36 +796,28 @@ def register_callback(signal, callback):
         pyblish._registered_callbacks[signal] = [callback]
 
 
-def deregister_callback(callback):
+def deregister_callback(signal, callback):
+    """Deregister a callback
 
-    for signal in pyblish._registered_callbacks:
-        if callback in pyblish._registered_callbacks[signal]:
-            pyblish._registered_callbacks[signal].remove(callback)
+    Arguments:
+        signal (string): Name of signal to deregister the callback with.
+        callback (func): Function to execute when a signal is emitted.
+    """
 
-
-def deregister_callbacks(signal):
-
-    pyblish._registered_callbacks.pop(signal, None)
+    if callback in pyblish._registered_callbacks[signal]:
+        pyblish._registered_callbacks[signal].remove(callback)
 
 
 def deregister_all_callbacks():
+    """Deregisters all callback"""
 
     pyblish._registered_callbacks = dict()
 
 
 def registered_callbacks():
+    """Returns registered callbacks"""
 
     return pyblish._registered_callbacks
-
-
-def emit(signal, **kwargs):
-
-    if signal in pyblish._registered_callbacks:
-        for callback in pyblish._registered_callbacks[signal]:
-            try:
-                callback(**kwargs)
-            except Exception as e:
-                log.error(e)
 
 
 def register_plugin(plugin):
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -482,6 +482,8 @@ def process(plugin, context, instance=None, action=None):
             provider.invoke(runner)
             result["success"] = True
     except Exception as error:
+        emit("pluginFailed", plugin=plugin, context=context, instance=instance,
+             error=error)
         pyblish.lib.extract_traceback(error)
         result["error"] = error
 
@@ -778,6 +780,46 @@ def current_host():
     """
 
     return pyblish._registered_hosts[-1] or "unknown"
+
+
+def register_callback(signal, callback):
+
+    if signal in pyblish._registered_callbacks:
+        pyblish._registered_callbacks[signal].append(callback)
+    else:
+        pyblish._registered_callbacks[signal] = [callback]
+
+
+def deregister_callback(callback):
+
+    for signal in pyblish._registered_callbacks:
+        if callback in pyblish._registered_callbacks[signal]:
+            pyblish._registered_callbacks[signal].remove(callback)
+
+
+def deregister_callbacks(signal):
+
+    pyblish._registered_callbacks.pop(signal, None)
+
+
+def deregister_all_callbacks():
+
+    pyblish._registered_callbacks = dict()
+
+
+def registered_callbacks():
+
+    return pyblish._registered_callbacks
+
+
+def emit(signal, **kwargs):
+
+    if signal in pyblish._registered_callbacks:
+        for callback in pyblish._registered_callbacks[signal]:
+            try:
+                callback(**kwargs)
+            except Exception as e:
+                log.error(e)
 
 
 def register_plugin(plugin):
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -8,9 +8,6 @@ options, such as ".exe" or ".bat".
 In this system, the predicate is whether or not a fname starts
 with "validate" and ends with ".py"
 
-Attributes:
-    patterns: Regular expressions used for lookup of plugins.
-
 """
 
 # Standard library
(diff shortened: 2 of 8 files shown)


mottosso:
Thanks @tokejepsen, having a proper look at this tonight.

mottosso:
Adding some additional info for completeness.

### Goal

Handle GUI events - such as the user toggling an Instance on/off.

### Usage
- **Step 1** - Create handler for the event.
  
  ``` python
  def toggle_handler(item, data_item, new_value, old_value):
      print("%s.data[\"%s\"] = %s (was %s)" % (item, data_item, new_value, old_value))
  ```
- **Step 2**  - Register
  
  ``` python
  import pyblish.api
  pyblish.api.register_handler(toggle_handler)
  ```

Now whenever an item is toggled, it would print:

```
myInstance.data["isToggled"] = False (was True)
```

Given that an instance named "myInstance" was toggled.

### Full Example

The below runs in Maya as-is, and produces a toggleable Instance and Plug-in.

``` python
import pyblish.api

def toggle_handler(item, data_item, new_value, old_value):
    print("%s.data[\"%s\"] = %s (was %s)" % (item, data_item, new_value, old_value))

pyblish.api.register_handler(toggle_handler)

class Collector(pyblish.api.Collector):
    def process(self, context):
        context.create_instance("myInstance")

class Validator(pyblish.api.Validator):
    optional = True

import pyblish_maya
pyblish_maya.show()
```

### Feedback

There are a few things that pop out to me.
1. Handlers can only be assigned to a single event
2. Handlers can be registered multiple times
3. There is no way of deregistering handlers
4. `itemToggled` might better be `publish`

For (1), I would suggest including an identifier towards the even a handler is meant to listen for. Think of it is the equivalent `button.pressed.connect(my_handler)`, except in this case it might look like `pyblish.api.register_handler("buttonPressed", my_handler)`.

Alternatively, we could got full-Qt.

``` python
pyblish.api.item_toggled.connect(my_handler)
```

Maybe it would probably be a better fit for pyblish-qml, seeing as the even is coming from there?

``` python
pyblish_qml.item_toggled.connect(my_handler)
```

Or maybe it's suited to the core Pyblish, in case the event is triggered from somewhere else, like in headless mode? Would that be interesting?

I'd imagine a similar interface for other events too.

``` python
pyblish.api.register_handler("pluginFailed", handle_failed)
pyblish.api.register_handler("finished", print_report)
pyblish.api.register_handler("validated", handle_validated)
pyblish.api.register_handler("extracted", handle_extracted)
pyblish.api.register_handler("tabChanged", handle_extracted)
```

Where some are GUI-only, some are relevant to headless publishing and some to both. With this interface, the registration towards an even that doesn't exist isn't a big deal, like if there is a handler for the GUI, but the GUI isn't running. Or if an even is handled that isn't present because the version of Pyblish is a few versions old.

For (2), I would make it a dictionary, with a key of the function name.

For (3), a simple `deregister_handler()` could do the trick.

For (4), `isToggled` is unique to the GUI. It might be more intuitive to handle `publish` as that is the data already in the Instance that changes.

And finally, as we are talking about events in the context of Pyblish, I'll include this (slightly related) link here.
- https://github.com/pyblish/pyblish/issues/143

tokejepsen:
> Or maybe it's suited to the core Pyblish, in case the event is triggered from somewhere else, like in headless mode? Would that be interesting?

This does sound a lot more useful than just GUI related events. GUI events are limited (right now), so expanding it to operate on all events in the Pyblish framework sounds great.

Currently when toggling we explicitly call the handler([here](https://github.com/tokejepsen/pyblish-qml/blob/b6f11235fc3d8450af4961dc83f4cd27a827cbcf/pyblish_qml/control.py#L457-L458) > [here](https://github.com/tokejepsen/pyblish-rpc/blob/a62ae6601d64b1d70de321b9e0de2f04a27b0b62/pyblish_rpc/client.py#L83-L84) > [here](https://github.com/tokejepsen/pyblish-rpc/blob/a62ae6601d64b1d70de321b9e0de2f04a27b0b62/pyblish_rpc/service.py#L127-L143)), meaning that we would need to explicitly call the handlers from all events we need/want. I don't know whether there are cleverer solution to this?

mottosso:
> Currently when toggling we explicitly call the handler(here > here > here), meaning that we would need to explicitly call the handlers from all events we need/want. I don't know whether there are cleverer solution to this?

No I think this is precisely how it should work. Only the emitter knows when anything is emitted, and it's good to be explicit about it. It means we could potentially emit the same signal from multiple places, like if an instance got toggled by other means (such as in toggling them via the section label, and if that wasn't already triggering the same function in this case) or to mock or fake signals that aren't necessarily coming from where they say their coming from.

Bottom line, this is a clean, lean and efficient implementation. Good work. :)

About the other points, what are you thoughts?

tokejepsen:
> About the other points, what are you thoughts?

They all make sense, and I'll have a look at them for the next iteration:)

mottosso:
> meaning that we would need to explicitly call the handlers from all events we need/want.

Re-reading this, about not so much emitting signals (which is as I said, perfect as it is right now) but actually triggering the response; yes, we might want to implement a "headquarters" in which the actual handlers are being called.

``` python
pyblish.api.emit("itemToggled", {"some": "data"})
self.host.emit("itemToggled", {"some": "data"})
```

The handlers would then need to be registered where those guys could access them, such that they know what to trigger, which I think is already the case.

Edit: Sorry, to clarify; the fact that a signal is emitted from the actual place it is made is _good_. It makes sense, and shouldn't change. That was what I was commenting on in the previous post. The "headquarters" would simply be adding potential extra methods of Proxy, like [this one](https://github.com/tokejepsen/pyblish-rpc/blob/a62ae6601d64b1d70de321b9e0de2f04a27b0b62/pyblish_rpc/client.py#L83-L84), together into a single method/function somewhere that could handle the arbitrary "key", e.g. `itemToggled`, and call the appropriate handlers registered to that value.

Does that make sense? :S

mottosso:
Added clarification.

tokejepsen:
There are a new PR across pyblish, pyblish-qml and pyblish-rpc.

mottosso:
:+1: Great work.

You'll notice some errors on [pyblish-rpc](https://github.com/pyblish/pyblish-rpc/pull/10), but don't fret it. It's because the tests are using the current version of Pyblish, which doesn't have the new functionality yet.

Once we've merged the in the right order, starting with pyblish, those tests will make more sense.

tokejepsen:
I've been thinking more about signals. How about we namespace the signals?

By namespacing the signals we could have the same signal from pyblish-qml and pyblish without clashing. This would also make it easier to identify where the signal are coming from.

mottosso:
That's a really cool idea. Are you thinking something along these lines?

``` python
pyblish.api.emit("pyblish-qml:itemToggled", my_handler)
```

I would probably wait with it, at least until we've seen how many signals we end up with. No need to add namespaces to 5-6 signals, or we'd just end up cluttering. Once we've seen and used these things in action for a while, it will become more obvious whether an additional level of organisation is a good idea.

tokejepsen:
> Are you thinking something along these lines?

Yup, pretty much😀

When we have this merged, I'll be implementing it in our pipeline.

mottosso:
> When we have this merged

Just for clarity, are you waiting for me to comment on anything or to merge this, or do you have things left to do? We will at least need to pass the coveralls by adding tests for the new functionality.

tokejepsen:
> Just for clarity, are you waiting for me to comment on anything or to merge this, or do you have things left to do?

Was waiting on a merge:) Don't know how the coveralls work, so will look at it tomorrow to get it passed.

mottosso:
Ah! Ok, having a closer look at the code then. :)

pyblish/plugin.py
```diff
@@ -780,6 +782,46 @@ def current_host():
...
 
+def register_callback(signal, callback):
+
+    if signal in pyblish._registered_callbacks:
+        pyblish._registered_callbacks[signal].append(callback)
+    else:
+        pyblish._registered_callbacks[signal] = [callback]
+
+
+def deregister_callback(callback):
```
mottosso: Hm, there's probably no need for both `deregister_callback` and `deregister_callbacks`. Sure, it will reduce typing, but increase maintenance. Remember, we will need to keep both of these functions forever and ever. The less of that the better!

Additionally, I would make the signature identical to the registration. Just so we don't need to remember two signatures.

``` python
deregister_callback("myEvent", my_callback)
```

pyblish/plugin.py
```diff
@@ -780,6 +782,46 @@ def current_host():
...
+
+    pyblish._registered_callbacks = dict()
+
+
+def registered_callbacks():
+
+    return pyblish._registered_callbacks
+
+
+def emit(signal, **kwargs):
```
mottosso: I think `emit()` fits better outside of `plugin.py`, perhaps in `lib.py` as it is not really related to anything in particular and is more of a utility function?

pyblish/plugin.py
```diff
@@ -780,6 +782,46 @@ def current_host():
...
+
+    if signal in pyblish._registered_callbacks:
```
mottosso: How about this?

``` python
for callback in pyblish._registered_callbacks.get(signal, []):
  try:
    callback(**kwargs)
 except Exception as e:
   log.error(e)
```

That way, nothing happens if there aren't any of those signals, and we reduce one line and one level of indentation.

Also, what _should_ happen when a callback fails? I would not silence it. I would let it roar.

``` python
for callback in pyblish._registered_callbacks.get(signal, []):
  try:
    callback(**kwargs)
 except Exception as e:
   # Print full exception message, but don't break the loop
   traceback.print_exc(e)
```

This way you still get information on which line and file the error occurred, as opposed to just the final message.

pyblish/plugin.py
```diff
@@ -780,6 +782,46 @@ def current_host():
     return pyblish._registered_hosts[-1] or "unknown"
 
 
+def register_callback(signal, callback):
+
```
mottosso, line 791: Don't forget docstrings to new functions.

mottosso:
> Don't know how the coveralls work

Coveralls will run all tests, and see what lines of code were run during that time.

What you'll need to do, is look inside of `/tests` and see how a test is typically written, and then write a series of tests that exercises the new functionality.

A test should try and break the code, find ways of making the functionality behave in unexpected ways. Like registering a function of the same name multiple times, and deregistering it afterwards. Does it still work?

Once you've made a test, run `run_testsuite.py`. It will run that all tests and give you a report of how it went. When you're confident that you've captured all aspects of the functionality, run `run_coverage.py` it will produce a `.html` document with a report on which lines of code were exercised during the tests.

pyblish/plugin.py
tokejepsen: How about `log.error(traceback.format_exc())`, to get the full stack?

mottosso: Why use `log` here?

tokejepsen: I thought it was best to keep all the records together in the same log?

mottosso: Hmm, are you thinking about records stored in the `result` of each process? Do they get captured here? I mean, do they appear in the GUI?

tokejepsen: > Hmm, are you thinking about records stored in the result of each process?

No, it was more as a convention. I don't mind either way, just trying to follow suit in the code.

mottosso: Ah.. I can't tell whether it'd make a difference. Maybe it will, but let's leave it for when it becomes obvious.

tokejepsen:
Passed the coverage! Woop woop:)

mottosso:
Nice! Having a closer look asap, but from first glance, it looks good!

pyblish/lib.py
```diff
@@ -349,3 +351,27 @@ def _resolve_name(name, package, level):
...
+
+            pyblish.api.register_callback("someSignal", myCallback)
+            emit("someSignal", data={"something": "cool"})
+    """
+    for callback in pyblish._registered_callbacks.get(signal, []):
+        try:
+            callback(**kwargs)
+        except Exception as e:
+            traceback.print_exc(e)
+            raise Exception(e)
```
mottosso: You added a `raise` here, that's problematic. It means the loop will stop once it throws the first exception, meaning other callbacks won't get triggered due to one of them being bad.

tokejepsen: Yeah, that was to be able to test the code properly. Couldn't figure out how to trigger coveralls with just printing.

mottosso: Ah, I see. One trick is to throw an exception in the test while building it, like `assert False`, just after it prints. That way, it will throw your exception, and show you what got printed before it. Once you've seen that it prints the right things, you can remove the temp exception.

pyblish/plugin.py
```diff
@@ -780,6 +782,44 @@ def current_host():
...
+    """
+
+    if callback in pyblish._registered_callbacks[signal]:
+        pyblish._registered_callbacks[signal].remove(callback)
+
+
+def deregister_all_callbacks():
+    """Deregisters all callback"""
+
+    pyblish._registered_callbacks = dict()
```
mottosso: Rather than creating a new dictionary and re-assigning it, it can be good to call `.clear()` instead. That way, anyone holding a reference to that dictionary won't be invalidated.

pyblish/lib.py
```diff
@@ -349,3 +351,27 @@ def _resolve_name(name, package, level):
...
+def emit(signal, **kwargs):
+    """Searches through the registered callbacks and executes any callback that
+    matches the signal.
+    The keyword arguments (kwargs) are passed directly through from the emitter
+    to the callbacks.
+
+        Arguments:
+            signal (string): Name of signal emitted
+
+        Example:
```
mottosso: When you make an example using `>>>` in a docstring, it qualifies as a "doctest". Meaning it'll get run during testing and what looks like the output in the example will be the assertion it makes. In this case, the output is expected to be `{'something': 'cool'}` so if you put that underneath the `>>>` like it would have appeared in a terminal, it will indicate failure during testing if it ever goes bad. :)
- https://docs.python.org/2/library/doctest.html

mottosso: For example:

``` python
"""
Example:
    >>> import sys
    >>> from .plugin import register_callback
    >>> register_callback("mysignal", lambda data: sys.stdout.write(data))
    >>> emit("mysignal", data={"something": "cool"})
    {'something': 'cool'}
"""
```

You can test the test, with `doctest.testmod()` from another file or terminal.

``` python
>>> import doctest
>>> import pyblish.lib
>>> doctest.testmod(pyblish.lib)
```

mottosso:
Hey @tokejepsen are you working on this? I've found a few other issues with the code in general that I'd like to address, mainly about using relative imports, that would have a small effect on many place in the code. It would be best to merge this before doing that, or if you haven't started yet, I can send a PR to you.

tokejepsen:
> Hey @tokejepsen are you working on this?

Nearly done, just need to address the docstring.

mottosso:
Cool

mottosso:
I noticed that when the doctest runs, it will register a callback. That callback will then survive until all testing is done. It can be a good idea to deregister all callbacks in tests that rely on there not being any callbacks.

mottosso:
If you look at some of the other tests, they have a `with_setup` decorator attached. That decorator is meant to "flush" the memory of previous tests, such that the current test gets a clean slate. That could work for you as well, you will just need to append flushing of callbacks to `lib.setup_empty`.

tokejepsen:
> I noticed that when the doctest runs, it will register a callback. That callback will then survive until all testing is done. It can be a good idea to deregister all callbacks in tests that rely on there not being any callbacks.

Yeah, just noticed that as well. Thanks for the advice on direction, I would have been lost:)

tokejepsen:
Alright, my pen is down:) I'll await your PR.

mottosso:
Sweet :) Sending shortly.

mottosso:
Success!

tokejepsen:
Now to the rest of the repos:)

mottosso:
Yeah! As pyblish-qml depends on pyblish-rpc, we'll have to handle that first.

Issue #239, opened 2016-01-21
Iterator


mottosso:
### Goal

Delegate the responsibility of choosing next pair of plug-in/instance to controller. (And replace [logic.process](https://github.com/pyblish/pyblish/blob/65327d0eabedc7d7da5023097ca51b39da9e0e35/pyblish/logic.py#L51))
### Motivation

Currently, there is a master function within the core library which provides the controllers - Pyblish QML, pyblish.util and Pyblish CLI - with utility and information about how to process and what happened once processing completed.

This has worked great, as it meant the primary logic - the logic determining which plug-in and how to process it and when to stop etc. - could be centralised.

It did have one minor inconvenience however, which was that the controllers were given so much that they didn't have enough influence over this logic to actually determine which pair would be the next pair to process. Would it be `{plug-in1: instance1}`, or `{plug-in-1: instance2}`? There was no way of knowing. Instead, the controller delivered _all_ plug-ins and _all_ instances to this one central function and _it_ decided how and when to do things.

This side-effect was noticeable in the GUI as an inaccurate representation of the currently running plug-in.
### Implementation

Apart from the above minor inconvenience, the delegation and separation of responsibility works great. So how do we extinguish the inconvenience?

Via an "iterator".

``` python
# Before
for result in logic.process(plugins, instances):
  # visualise result

# With Iterator
for plugin, instance in Iterator(plugins, instances):
  result = plugin.process(plugin, instance)
  # visualise result
```

As in the example, the solution delegates more responsibility to a controller, and that inevitably means more code must be written within both Pyblish QML, `pyblish.util`, Pyblish CLI and any future interface. But I feel this is unavoidable. The interface should provide this level of control and I can't envision a more optimal route in the current.
### Summary

The Iterator fully replaces [logic.process](https://github.com/pyblish/pyblish/blob/65327d0eabedc7d7da5023097ca51b39da9e0e35/pyblish/logic.py#L51) and introduces more code into each interface, but the gain is great and enables a great deal of necessary control.

PR #240, opened 2016-01-21
Added Iterator

mottosso:
Implements #239


mottosso:
Worth noting is that this is a _fundamental_ change to logic throughout Pyblish.

I'm not expecting anything to break as the behaviour from the previous implementation has been re-implemented and corner cases dealt with and all tests pass, but do keep an eye out for odd behaviour. Especially involving triggering of plug-ins that only depend on the Context, versus those that only depend on the Instance.

PR #244, opened 2016-02-03
- fix for pyblish-qml/#125

tokejepsen:


mottosso:
Remind me again why this was necessary? Isn't it enough to handle this in QML?

tokejepsen:
> Remind me again why this was necessary? Isn't it enough to handle this in QML?

You could, but that would mean that you had to accommodate for the key not existing in the plugins.

``` python
if instance.data.get("publish", True):
    print "Instance is publishable"
```

If we have the attribute on the instance by default, it would make more sense for new people as you don't have to know the inner workings of pyblish-qml, that it adds it on-demand.

``` python
if instance.data['publish']:
    print "Instance is publishable"
```

Are you trying to keep the data members to the minimum in the base code?

mottosso:
> Are you trying to keep the data members to the minimum in the base code?

Yes, but what do you think, is it worth it?

I could also imagine it being useful to have plentiful of default data members, sort of like how Maya has all of those default nodes. Maybe the constructor of the instance is a good place for those after all?

tokejepsen:
> is it worth it?

I think in this case, its quite confusing for a user when we instruct them that untoggling the instance by default in the GUI is done through the `publish` data member; http://forums.pyblish.com/t/solved-un-toggle-instance-by-default/91

> Maybe the constructor of the instance is a good place for those after all?

Yeah, definitely. What other default data members are you thinking about?

mottosso:
> Yeah, definitely. What other default data members are you thinking about?

I'm thinking there would be many defaults for the many disparate parts of Pyblish, not just publishing, and not just the GUI. I would want those defaults to reside within the corresponding project, as opposed to pollute the base.

I'm worried about there being loads of defaults related to some project that most aren't even using, like the pyblish-ci, or pyblish-tray. At the moment, those defaults are meant as part of the documentation for those modules.

tokejepsen:
> I'm worried about there being loads of defaults related to some project that most aren't even using, like the pyblish-ci, or pyblish-tray. At the moment, those defaults are meant as part of the documentation for those modules.

Good point, I'll see about adding the `publish` data member when using pyblish-qml.

mottosso:
> Good point, I'll see about adding the publish data member when using pyblish-qml.

Ok, let's make it a point from now on to document defaults in the corresponding projects. Adding this to the Developer Guide as well.

mottosso:
About the fail of the tests on this, it's because the name of the project has changed, but it isn't yet in your repo, nor in this base repo.

We'll ignore it for now, as I'll soon merge my fork which will resolve that. Just make sure you've tested whatever has been added or changed.

tokejepsen:
> We'll ignore it for now, as I'll soon merge my fork which will resolve that. Just make sure you've tested whatever has been added or changed.

Fairly confident that I can make it all happen in pyblish-qml, so will be closing this when I do.

mottosso:
Cool.

PR #263, opened 2016-04-01
Fix #250

### LINKED ISSUE #250: ContextPlugin running when instance of family not present
### Problem

A `ContextPlugin` with `families` specified will run when there's no instance with that family.
### Expected

I'd expect anything with `families` specified should only run when at least one instance is available with at least one of those families.

As such the above should only run if there are instances present of family `rig` and/or `model`. When those are present it should run only once, like any `ContextPlugin`.
### To reproduce

``` python
import pyblish.api
import pyblish.util


class CollectInstanceTest(pyblish.api.ContextPlugin):

    def process(self, context):
        print("Running %s" % type(self))
        instance = context.create_instance("testInstance")
        instance.data["family"] = "thisFamily"


class ValidateContextTest(pyblish.api.ContextPlugin):
    families = ["otherFamily"]

    def process(self, context):
        print("Running %s" % type(self))


pyblish.api.register_plugin(CollectInstanceTest)
pyblish.api.register_plugin(ValidateContextTest)

pyblish.util.publish()
# Running <class '__main__.CollectInstanceTest'>
# Running <class '__main__.ValidateContextTest'> <-- This should not be called
```
### Also see

This was also started as discussion on the forums [here](http://forums.pyblish.com/t/solved-contextplugin-running-when-instance-of-family-not-present/202). It seems somewhat related to the fixed #243 since it also required an instance to exists (operating on instance). In this issue we're not operating on an instance, but on the context (run only once).

This issue is still present in `1.3.2`

tokejepsen:


mottosso:
Could you have a look at the breaking test?

tokejepsen:
The breaking test seems to be against what we are trying to solve.

The test is [here](https://github.com/pyblish/pyblish-base/blob/62d767f4386d33e7cb18f544086ac128af02c1bc/tests/pre13/test_logic.py#L145), and fails [here](https://github.com/pyblish/pyblish-base/blob/62d767f4386d33e7cb18f544086ac128af02c1bc/tests/pre13/test_logic.py#L174).
Test says that a plugin with a context argument, meaning a `ContextPlugin`, should still run even though no instances of the family specified is present. Here is the test isolated:

``` python
import pyblish.api
import pyblish.util


class SelectMany(pyblish.api.Selector):
    def process(self, context):
        for name in ("A", "B", "C"):
            instance = context.create_instance(name)
            instance.set_data("family", "myFamily")


class ValidateContext(pyblish.api.Validator):
    families = ["NOT_EXIST"]

    def process(self, context):
        print "I shouldn't run"

pyblish.util.publish(plugins=[SelectMany, ValidateContext])
```

tokejepsen:
BTW didn't see this failure before because the `run_testsuite.py` I executed was in the wrong environment.

tokejepsen:
Does this mean we have to change the test?

mottosso:
Sorry I haven't had a look at this yet. It's on my todo.

mottosso:
Closing this; discussion [here](https://github.com/pyblish/pyblish-base/issues/250).

Issue #266, opened 2016-04-26
Instance unique ids


BigRoy:
# Goal

To provide unique ids for instances so that instances with the same names can be identified uniquely in the Context.

This is also useful to simplify work on the GUI.
# Changes

Currently the `instance.id` is a property using the instance's name, as such instances with the same name don't have a unique id. The change would be to store a unique id (eg. `uuid.uuid4()`) attribute on the instance during initialization.

**Note:**
This change also means that the instance can't be retrieved from the Context anymore by name. This could be a potentially backwards incompatible change if one was using that (convenient) behavior. For example many tests were written this way.

PR #267, opened 2016-04-26
Implement #266

BigRoy:
- Instances have a unique id generated with uuid.uuid4()

### DIFF
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -56,19 +56,19 @@ def test_instance_equality():
     inst3 = pyblish.plugin.Instance('Test2')
 
     assert inst1 != inst2
-    assert inst2 == inst3
+    assert inst2 != inst3
 
 
 def test_context_itemgetter():
     """Context.get() works"""
     context = pyblish.api.Context()
-    context.create_instance("MyInstanceA")
-    context.create_instance("MyInstanceB")
+    instanceA = context.create_instance("MyInstanceA")
+    instanceB = context.create_instance("MyInstanceB")
 
-    assert context["MyInstanceA"].name == "MyInstanceA"
-    assert context["MyInstanceB"].name == "MyInstanceB"
-    assert context.get("MyInstanceA").name == "MyInstanceA"
-    assert context.get("MyInstanceB").name == "MyInstanceB"
+    assert context[instanceA.id].name == "MyInstanceA"
+    assert context[instanceB.id].name == "MyInstanceB"
+    assert context.get(instanceA.id).name == "MyInstanceA"
+    assert context.get(instanceB.id).name == "MyInstanceB"
     assert context[0].name == "MyInstanceA"
     assert context[1].name == "MyInstanceB"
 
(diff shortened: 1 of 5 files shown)


BigRoy:
This will likely need some changes in `pyblish-qml` to work correctly.

tests/test_context.py
```diff
@@ -56,19 +56,19 @@ def test_instance_equality():
...
-    context.create_instance("MyInstanceA")
-    context.create_instance("MyInstanceB")
+    instanceA = context.create_instance("MyInstanceA")
+    instanceB = context.create_instance("MyInstanceB")
 
-    assert context["MyInstanceA"].name == "MyInstanceA"
-    assert context["MyInstanceB"].name == "MyInstanceB"
-    assert context.get("MyInstanceA").name == "MyInstanceA"
-    assert context.get("MyInstanceB").name == "MyInstanceB"
+    assert context[instanceA.id].name == "MyInstanceA"
```
mottosso, line 68: This is nice!

mottosso:
Make sure you get the str-representation of the uuid, we'll need it plain and serialisable.

``` python
>>> import pyblish.api
>>> context = pyblish.api.Context()
>>> instance = context.create_instance("MyInstance")
>>> instance.id
UUID('17840082-b0f6-4eab-8318-34720ff8a3f7')  # <- Should be str
```

BigRoy:
> Make sure you get the str-representation of the uuid, we'll need it plain and serialisable.

Thanks, totally missed that. Changed that with https://github.com/pyblish/pyblish-base/pull/267/commits/af70d41c9cc3d43d7e9e88c4a9c081016768dcbd

mottosso:
+0.004% coverage increase, I like it. :)

mottosso:
Great work, Roy. Having a closer look now.

mottosso:
Ok, not sure we're pushing our luck, or whether we should go for consistency right away, but what do you think about adding `id` to Plugins and Context as well?

BigRoy:
> Ok, not sure we're pushing our luck, or whether we should go for consistency right away, but what do you think about adding id to Plugins and Context as well?

Yes, exactly. Problem I have is figuring out where and when to add those since Plugins aren't always initialized and as such when does the uuid get created? Upon registering?

For `Context` I think it's the same as with `Instance` that it does get initialized and could just be added to `__init__`?

mottosso:
> For Context I think it's the same as with Instance that it does get initialized and could just be added to **init**?

I think so too.

> Upon registering?

Could do, or rather in the discover() function. An alternative is the [MetaPlugin](https://github.com/pyblish/pyblish-base/blob/62d767f4386d33e7cb18f544086ac128af02c1bc/pyblish/plugin.py#L174).

discover() is pretty clean, but there are cases where one might call a plug-in directly, or pass it directly to another function, and in those cases it would not be getting an id. :(

The MetaPlugin runs on module import and would be more bulletproof.

BigRoy:
MetaPlugin could work fine I guess.

Not sure how that works with PRs, but can you actually take over from here? (Have to run off and busy for a day or so)

mottosso:
> Can you take over from here?

Can do!

mottosso:
Continued in #264

Issue #234, opened 2015-11-20
Python 3


mottosso:
### Goal

Support Python 3.
### Motivation

The transition from 2-3 is inevitable, but the more pressing issue lies in the fact that the user interfaces are written using PyQt5 which is currently compiled individually for each platform and update to Qt.

If Pyblish were to run on both Python 2 and 3, it could benefit from PyQt5 being compiled and distributed by others, on any platform, which would simplify deployment tremendously.

Also automated testing would be simplified, as PyQt5 is available to major distributions including Travis-CI.
### Implementation

In essence, re-factor Python 2-only functionality into that which is compatible with both 2 and 3.

mottosso:
This is ongoing. yaml, nose and coverage were Python 2-only libraries that were recently shed. The next immediate issues are related to subprocess management and the plug-in discovery mechanism, both of which rely on behaviour specific to Python 2.

mottosso:
This is done!

PR #274, opened 2016-05-05
Python 3 Support

mottosso:
This is an ongoing Pull Request for work to implement support for Python 3.

It includes `six` and a few patches to the overall project, but a large number of tests still fail. Most of these issues are expected to be cosmetic, with the exception of the `discover()` dependency on `execfile` which has been removed from Python 3.

Motivation for this can be found in issue #234.

What you can do, is have a look at what fails currently and what can be done to resolve it. The goal is for Pyblish to be written in Python 3, with Python 2 support, rather than the other way around. So when monkeypatching is required, it is Python 2 which must be monkeypatched to resemble Python 3.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,4 +11,4 @@ docs/_*
 build
 playground.py
 *.sublime-*
-__*
+__pycache__
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ docs/_*
 build
 playground.py
 *.sublime-*
+__*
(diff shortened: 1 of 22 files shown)


mottosso:
Here's an example of easy fixes.
- https://travis-ci.org/pyblish/pyblish-base/jobs/128055483#L869
- https://travis-ci.org/pyblish/pyblish-base/jobs/128055483#L846

Somewhat challenging
- https://travis-ci.org/pyblish/pyblish-base/jobs/128055483#L765

More challenging
- https://travis-ci.org/pyblish/pyblish-base/jobs/128055483#L657

mottosso:
Well, that didn't take as long as I had expected. :)

mottosso:
This is happening!

PR #279, opened 2016-06-22
register_gui()

mottosso:
### Goal

Enable use of multiple GUIs and eliminate dependency on pyblish-integration and pyblish-qml from host integration packages, like pyblish-maya.
### Implementation

A new function is introduced, where a developer may build a GUI and register it with Pyblish. An integration may then query registered GUIs and attempt to open the most suitable one.
- http://api.pyblish.com/register_gui.html

Also see:
- https://github.com/pyblish/pyblish-qml/pull/204
- https://github.com/pyblish/pyblish-maya/pull/57

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,24 @@
+### Contributing to Qt
+
+To contribute, fork this project and submit a pull-request.
+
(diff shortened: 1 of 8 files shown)


mottosso:
Note to self, contribution guidelines mention Qt.py project.

Issue #283, opened 2016-07-06
Common processing loop


mottosso:
### Goal

Provide a common interface for graphical user interfaces as well as custom and/or complex processing loops.
### Overview

(1) `pyblish.plugin` provides (internal) low-level functionality towards Pyblish. (2) `pyblish.util` on the other hand offers a (public) high-level interface for common tasks. (3) `pyblish-qml` and `pyblish-lite` then provides the artist interface in the form of graphical user interfaces.

What's missing here is (4) low-level, but public interface for developers to build their own complex tools and solve complex requirements.

It should be simple enough for general use, yet complex enough to accommodate for a fully featured graphical user interface.
### Motivation

pyblish.util, pyblish-qml and pyblish-lite currently maintain their own unique processing loops. The benefit of this has been that changes in one loop does not negatively affect another which has made it welcoming to experiment with new approaches to processing in search of the most optimal one.

A disadvantage to such an approach however is that maintenance is multiplied by the number of implementations. Tests are implemented multiple times, and odds are one behaves slightly different then the next.

This issue is an attempt to solve this.
### Demotivation

I've tried this before.

`pyblish.logic.process` (now [deprecated](https://github.com/pyblish/pyblish-base/blob/master/pyblish/compat.py#L147)) provided an attempt at bridging the gaps between GUI, CLI and Python interface. The problem was subtle; too complex for general use yet too simple for the requirements of any GUI.
### Implementation

`engine.py` will be in the family of `util.py`; that is, a completely standalone entity with no other dependencies than pyblish itself, and pyblish-base not depending on anything within it. A top-level dependency. `util.py` may eventually transition to using this.

The recent success in implementing the processing loop for pyblish-lite spawned the idea that perhaps such a loop could be made available on a more general basis and become shared across all interfaces.

An implementation must enable:
1. Scripting
2. Command-line use
3. Graphical user interfaces

And as such must be:
1. End-user friendly
2. Asynchronous
##### End-user friendly

At the end of the day, I'd like to enable the end-user with the power to implement their own complex requirements using `engine.py`. For example, bespoke graphical user interfaces, such as the one [developed at CAOZ](https://github.com/pyblish/pyblish-caozui), or unique one-off tasks, such as [wedges in Maya at Colorbleed](https://github.com/pyblish/pyblish-caozui).
##### Asynchronous

It must be asynchronous, a GUI should enable use even though it may be busy processing. This is because processing is expected to take anywhere between a few microseconds to minutes.

[...]

PR #284, opened 2016-07-06
Initial version of #283

mottosso:
Implementation of #283

See [`test_engine.py`](https://github.com/mottosso/pyblish-base/blob/9554bee6ad3919e5cc552b5f70ed0b8119cf392f/tests/test_engine.py) and [pyblish-lite.window.py](https://github.com/mottosso/pyblish-lite/blob/7defe969ff2988875bbd3b7fae935a23772e4d47/pyblish_lite/window.py#L67) for examples.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,13 @@ pyblish Changelog
 
 This contains all major version changes between pyblish releases.
 
+Version 1.4.2
+-------------
+
+- Implemented #283, common processing loop
+- Implemented #285, memory optimisation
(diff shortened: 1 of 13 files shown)


mottosso:
This is now up and running and very stable!

I'll continue integrating this into pyblish-lite, and later pyblish-qml. Once those are both using these mechanics, it'll be merged and enable others to build custom GUIs and tools that reap the benefits a common base and extensive testing.

This could be big!

mottosso:
Getting this to work in the multithreaded environment which is pyblish-qml shed some light on a few problems that stripped some if not most of the benefit provided by this feature.

It's at the back of my head currently, awaiting a moment of eureka.

Issue #282, opened 2016-07-04
pyblish.util.publish() ignores instance "publish" state.


BigRoy:
# Problem

The `publish` state of an instance is not used in `pyblish.util.publish()`. As such even when False it will process those instances.
# Solution

See [gitter discussion](https://gitter.im/pyblish/pyblish?at=577a8cd14fd381610991f991)

This:

``` python
    # Second pass, the remainder
    for Plugin, instance in logic.Iterator(plugins, context):
        state["nextOrder"] = Plugin.order
        ...
```

Should become:

``` python
    # Second pass, the remainder
    for Plugin, instance in logic.Iterator(plugins, context):
        if not instance.data.get("publish", True):
            continue

        state["nextOrder"] = Plugin.order
```

mottosso:
We could have a look at how Lite or QML does it.

``` python
        for plug, instance in pyblish.logic.Iterator(plugins, context):
            if not plug.active:
                continue

            if instance is not None and instance.data.get("publish") is False:
                continue
```
- [pyblish-lite.control.py](https://github.com/pyblish/pyblish-lite/blob/3538a2fbe9191357ce72da2e5d4ff5dc5441c62a/pyblish_lite/control.py#L228)

There we have it!

mottosso:
Implemented.

PR #290, opened 2016-07-19
Implement #282

mottosso:
Also moved this responsibility into the Iterator, now no GUI will need to self-implement this such that all implementations will have a more aligned behavior.

Also added log messages at DEBUG level that can be enabled like so.

``` python
import logging
logging.getLogger("pyblish").setLevel(logging.DEBUG)
```

These will tell you when (1) a plug-in is skipped due to being `active=False` and (2) when an instance is skipped due to `data["publish"] = False`.

### DIFF
--- a/pyblish/logic.py
+++ b/pyblish/logic.py
@@ -1,10 +1,14 @@
 """Shared processing logic"""
 
 import sys
+import logging
 import traceback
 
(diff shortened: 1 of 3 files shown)


mottosso:
Going with this. The failure was on part of increased test-coverage, but I'll be loosening up this requirement from being this draconian about it.

PR #291, opened 2016-07-21
Added pluginProcessed event

pscadding:
Added a new event being emited to the plugin.__explicit_process

This will emit when ever a plugin is processed regardless of result

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -511,7 +511,7 @@ def __explicit_process(plugin, context, instance=None, action=None):
     context.data["results"].append(result)
 
     lib.emit("pluginProcessed", plugin=plugin, context=context,
-             instance=instance, results=result)
+             instance=instance, result=result)
 
     return result
 
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -440,9 +440,12 @@ def process(plugin, context, instance=None, action=None):
     """
 
     if issubclass(plugin, (ContextPlugin, InstancePlugin)):
-        return __explicit_process(plugin, context, instance, action)
+        result = __explicit_process(plugin, context, instance, action)
     else:
-        return __implicit_process(plugin, context, instance, action)
+        result = __implicit_process(plugin, context, instance, action)
+
+    lib.emit("pluginProcessed", result=result)
+    return result
 
 
 def __explicit_process(plugin, context, instance=None, action=None):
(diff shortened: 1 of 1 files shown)


mottosso:
Thanks for this @pscadding, I think it's an excellent feature.

Linking to the corresponding Forum thread.
- http://forums.pyblish.com/t/pluginprocessed-callback/264

About the change, could you make sure your commits are correctly authored to you? The username and email you use on GitHub must correspond to the one used to make the commits, otherwise they will become anonymous as you can see above.

Secondly, I don't think you need all of the keyword arguments, but instead only the `result=`, since it will in turn contain reference to the corresponding plugin, context and instance.

Thirdly, could you also make sure this applies to implicit plug-ins as well? You are welcome to put this signal in the main `process()` function if you wish.

Finally, there will need to be one or more tests to exercise all of what you expect of the feature.

pscadding:
Ah sorry about that I didn't notice that they had come through anonymous. hmmm, I'll have to look into that, we use bitbucket at work, and so I use my work email, but my Github was set to my personal. I'll look into it tomorrow.

Yeah your quite right actually I don't need the other keywords. 
I noted that it probably should go in the implicit, but wasn't sure if putting it in the explicit was the right place anyway. I sort of created the pullrequest knowing it wasn't finished but to use it as a holder for the commits. I probably should have branched it instead.

I've not done tests before, believe it or not, and its something I really should learn. I'll look into it.

How do I go about updating the api docs, or is that something you will do?

mottosso:
No problem at all, @pscadding! I'm expecting pull-requests to be work in progress, that's what makes these code review sessions so useful and why I gave you so many comments. Glad to hear you are excited about addressing them!

About updating the API docs, you are most welcome to do so.
- https://github.com/pyblish/apidocs

pscadding:
OK I know my most recent commit was anonymous but I think I have fixed it now, so all my future ones should be OK. I can refork and add the commit again if you would prefer.

pyblish/plugin.py
```diff
@@ -440,9 +440,12 @@ def process(plugin, context, instance=None, action=None):
     """
 
     if issubclass(plugin, (ContextPlugin, InstancePlugin)):
-        return __explicit_process(plugin, context, instance, action)
+        result = __explicit_process(plugin, context, instance, action)
     else:
-        return __implicit_process(plugin, context, instance, action)
+        result = __implicit_process(plugin, context, instance, action)
+
+    lib.emit("pluginProcessed", result=result)
```
mottosso, line 447: Nice solution. :)

mottosso:
> . I can refork and add the commit again if you would prefer.

Yes please.

Alternatively, you can modify your existing commits and force push to your existing fork. That would also update this PR and might be easier.

mottosso:
Continued here: https://github.com/pyblish/pyblish-base/pull/292

Issue #293, opened 2016-08-12
Rename Callback to Handler


mottosso:
### Goal

This is a purely cosmetic change, but I think also an important difference in mindset; **emit** signals and **handle** them.
### Motivation

A "callback" suggests running a function and having another function get "called back" once it's done. It's common in Javascript, which is asynchronous, for things like making a request and passing along a nameless function (i.e. lambda) that is to be run once it's done.

This isn't what is happening in Pyblish.

Instead, callbacks are more closely related to signals and slots of Qt, and I think we should treat them that way.
### Implementation

`register_callback` should instead be`register_handler`. Alternatively, perhaps more succintly as `pyblish.api.on("someEvent", my_handler)`

``` python
from pyblish import api

def on_toggled(instance, before, after):
  pass

# Register
api.on("instanceToggled", on_toggled)

# Deregister
del(on_toggled)

# Get
api.on("instanceToggled")  # == (on_toggled,)
```
### Backwards Compatibility

We can maintain both for the time being, mark the current one as deprecated, and let it be. It doesn't change what they do or how they are emitted and should be relatively painless to transition to.

tokejepsen:
> A "callback" suggests running a function and having another function get "called back" once it's done. 

Makes sense, I'm in:)

> pyblish.api.on("someEvent", my_handler)

Is this implemented elsewhere, or is this to follow more of a Qt signal/slots behaviour?

> We can maintain both for the time being, mark the current one as deprecated, and let it be. It doesn't change what they do or how they are emitted and should be relatively painless to transition to.

This is becoming a higher and higher priority; https://github.com/pyblish/pyblish-qml/issues/165 to be able to keep up with deprecation :) Maybe we should be able to search our code base for deprecation? (sorry about derailing the topic)

mottosso:
> Is this implemented elsewhere, or is this to follow more of a Qt signal/slots behaviour?

If you think about it, it's literally renaming the function `register_callback` to `on`. In that way, yes, I'd say it's already implemented. I just think it makes it easier to think about conceptually.

> Maybe we should be able to search our code base for deprecation?

Deprecated functionality is being tagged with a decorator currently, that spits out a warning when anything deprecated is being used. Now in theory, you could simply enable those warnings, and it'd tell you what and where you are using anything like that, but it hasn't seen much use and I'm pretty sure the base library is already using that internally (via `util.py`) for backwards compatibility, so it'll probably spit out a lot of stuff.

Here's an example.

``` cmd
[...]

PR #298, opened 2016-10-10
Rename Callback to Handler

mottosso:
This PR implements `on`, and renames all references to `callback` to `handler`.
- #293 (Rename Callback to Handler)

**Full backwards compatibility is preserved**

<br>

### Example

Technically, `on` is simply an alias for `register_handler`, previously known ad `register_callback`.

``` python
from pyblish import api, util

def on_published(context):
  print("Published!")

api.on("published", on_published)

util.publish()
# Published!
```

<br>

### Explicit Support for Graphical User Interfaces.

In addition, this PR implements "weak references" for all event handlers.

This shouldn't have much effect on your day to day use of Pyblish, but will make a difference in a few small corner cases.

For example, consider this.

``` python
count = {"#": 0}

class MyObject(object):
    def __init__(self):
        pyblish.api.on("mySignal", self.on_mysignal)

    def on_mysignal(self):
        count["#"] += 1

obj = MyObject()
pyblish.api.emit("mySignal")
assert_equals(count["#"], 1)

del(obj)
pyblish.api.emit("mySignal")
assert_equals(count["#"], 1)
```

As you would expect, once the object is deleted, the handler stops being called. Prior to this PR however, `count["#"]` equalled `2`.

In practice, this could cause issues when you build a GUI around signals, and register methods within your GUI and then create a new instance of this GUI.

The second instance would reconnect to these signals, leaving the old methods in place, resulting in two calls per method. For each instance of your GUI, the calls would stack up, and off you go.

This PR enables you to build user interfaces that harness the signals produces throughout Pyblish, without worrying about these difficult-to-debug problems.

### DIFF
--- a/pyblish/__init__.py
+++ b/pyblish/__init__.py
@@ -10,7 +10,7 @@ from .version import version, version_info, __version__
 
 
 _registered_paths = list()
-_registered_callbacks = dict()
+_registered_handlers = dict()
 _registered_plugins = dict()
 _registered_services = dict()
 _registered_test = dict()
@@ -24,7 +24,7 @@ __all__ = [
     "version_info",
     "__version__",
     "_registered_paths",
-    "_registered_callbacks",
+    "_registered_handlers",
     "_registered_plugins",
     "_registered_services",
     "_registered_test",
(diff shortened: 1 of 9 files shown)


mottosso:
This [fails on Python 3.3](https://ci.appveyor.com/project/mottosso/pyblish/build/211). :( Investigating..

Issue #301, opened 2016-10-17
Higher Orders not supported


tokejepsen:
Description

When using higher orders like 10 or 20, pyblish does not process correctly.

Short reproducible

This should output some print statements, but it doesn't;

``` python

import pyblish.api

items = ["john", "door"]

class CollectInstances(pyblish.api.ContextPlugin):
  order = 10

  def process(self, context):
    for item in items:
      context.create_instance(item)

class PrintInstances(pyblish.api.InstancePlugin):
  order = 20

  def process(self, instance):
    print("Instance is: %s" % instance)

pyblish.api.register_plugin(CollectInstances)
pyblish.api.register_plugin(PrintInstances)

import pyblish.util
pyblish.util.publish()
```

Lowering the orders does however output the expected results;

``` python

import pyblish.api

items = ["john", "door"]

class CollectInstances(pyblish.api.ContextPlugin):
  order = 0

  def process(self, context):
    for item in items:
      context.create_instance(item)

class PrintInstances(pyblish.api.InstancePlugin):
  order = 1

  def process(self, instance):
    print("Instance is: %s" % instance)

pyblish.api.register_plugin(CollectInstances)
pyblish.api.register_plugin(PrintInstances)

import pyblish.util
pyblish.util.publish()
```

mottosso:
Oh, that's bad! That means the learn by example will not have worked for a long time.

Having a look asap.

mottosso:
The problem lied in the documentation which has now been fixed.
- https://learn.pyblish.com/chapters/04-coordination.html

PR #302, opened 2016-10-18
fix for #301

tokejepsen:
Test code for context:

``` python
import pyblish.api

items = ["john", "door"]


class CollectInstances(pyblish.api.ContextPlugin):
    order = 10

    def process(self, context):
        for item in items:
            context.create_instance(item)


class PrintInstances(pyblish.api.InstancePlugin):
    order = 20

    def process(self, instance):
        print("Instance is: %s" % instance)

pyblish.api.register_plugin(CollectInstances)
pyblish.api.register_plugin(PrintInstances)

import pyblish.util
```

The issue was that when collection had been run, it was assumed that all plugins would have instances to process with. This discarded the wildcard `*` for all families, resulting in all instance plugins operating with `*` in their families would be removed from processing.

In theory this behavior is correct in that if there aren't any instances to operate on, the plugin shouldn't actually run. So this will fix it, but I don't know whether it should be fixed.

### DIFF
--- a/pyblish/util.py
+++ b/pyblish/util.py
@@ -54,7 +54,8 @@ def publish(context=None, plugins=None):
     # least one compatible instance.
     for Plugin in list(plugins):
         if Plugin.__instanceEnabled__:
-            if not logic.instances_by_plugin(context, Plugin):
+            if (not logic.instances_by_plugin(context, Plugin) and
+                "*" not in Plugin.families):
                 plugins.remove(Plugin)
 
     # Mutable state, used in Iterator


mottosso:
Ah, that's an interesting discovery and observation. I'm also not sure it should be fixed. I'll have a think about it.

mottosso:
Ok, this was a bug, not in the code but in the documentation. The fix was to have instances created only during collection - `order=0` - as opposed to `order=10`.

I've done a first pass at fixing it just now and most of it _should_ work well; but next I'll setup automated testing of those tests such that this can never happen again.

Thanks a lot for narrowing this down @tokejepsen, it's a relatively recent fault and so I hope it hasn't affected too many people.

PR #309, opened 2017-01-18
plugins_by_families now matches with algorithms

p4vv37:
Updated logic.plugins_by_families, now it uses the matching algorithm of
the plugin,
Updated test of plugins_by_families

### DIFF
--- a/pyblish/logic.py
+++ b/pyblish/logic.py
@@ -153,10 +153,26 @@ def plugins_by_families(plugins, families):
 
     """
 
+    algorithms = {
+        Intersection: lambda a, b: set(a).intersection(b),
+        Subset: lambda a, b: set(a).issubset(b),
+        Exact: lambda a, b: set(a) == set(b)
+    }
+
     compatible = list()
 
     for plugin in plugins:
-        if any(x in plugin.families for x in families + ["*"]):
+
+        if "*" in plugin.families:
+            compatible.append(plugin)
+            continue
+
+        algorithm = algorithms.get(plugin.match)
+
+        assert algorithm, ("Plug-in did not provide "
+                           "valid matching algorithm: %s" % plugin.match)
+
+        if algorithm(plugin.families, families):
             compatible.append(plugin)
 
     return compatible
--- a/pyblish/logic.py
+++ b/pyblish/logic.py
@@ -15,6 +15,11 @@ from .plugin import (
     Exact
 )
 
+_algorithms = {
+    Intersection: lambda a, b: set(a).intersection(b),
+    Subset: lambda a, b: set(a).issubset(b),
+    Exact: lambda a, b: set(a) == set(b)
+}
 
 log = logging.getLogger("pyblish.logic")
 
@@ -153,12 +158,6 @@ def plugins_by_families(plugins, families):
 
     """
 
-    algorithms = {
-        Intersection: lambda a, b: set(a).intersection(b),
-        Subset: lambda a, b: set(a).issubset(b),
-        Exact: lambda a, b: set(a) == set(b)
-    }
-
     compatible = list()
 
     for plugin in plugins:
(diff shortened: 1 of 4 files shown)


mottosso:
Thanks @p4vv37, would it be possible to provide an example of what didn't work before? A little runnable snippet of code.

p4vv37:
Sure, from changed tests:

    class ClassD(pyblish.api.Collector):
        families = ["a", "b"]
        match = api.Intersection

    class ClassE(pyblish.api.Collector):
        families = ["a", "b"]
        match = api.Subset

    class ClassF(pyblish.api.Collector):
        families = ["a", "b"]
        match = api.Exact

    assert pyblish.logic.plugins_by_families(
        [ClassD, ClassE, ClassF], ["a"]) == [ClassD]

    assert pyblish.logic.plugins_by_families(
        [ClassD, ClassE, ClassF], ["a", "b"]) == [ClassD, ClassE, ClassF]

    assert pyblish.logic.plugins_by_families(
        [ClassD, ClassE, ClassF], ["a", "b", "c"]) == [ClassD, ClassE]

p4vv37:
Sorry, I've just missclicked.

The snipplet of code that runs after change and did not run before it:


    class ClassD(pyblish.api.Collector):
        families = ["a", "b"]
        match = api.Intersection
 
    class ClassE(pyblish.api.Collector):
        families = ["a", "b"]
        match = api.Subset

    class ClassF(pyblish.api.Collector):
        families = ["a", "b"]
        match = api.Exact

    assert pyblish.logic.plugins_by_families(
        [ClassD, ClassE, ClassF], ["a"]) == [ClassD]

    assert pyblish.logic.plugins_by_families(
        [ClassD, ClassE, ClassF], ["a", "b"]) == [ClassD, ClassE, ClassF]

    assert pyblish.logic.plugins_by_families(
        [ClassD, ClassE, ClassF], ["a", "b", "c"]) == [ClassD, ClassE]


I've included it to the pyblish-base\tests\pre13\__init__.py

mottosso:
Cool, I'll have a closer look at this as soon as I can.

About the `preXX` folders and files, they are for tests of older versions of Pyblish; in this case, `pre13` means it includes tests for version 1.3 of Pyblish. They aren't anything that should ever change, as that would mean that newer functionality breaks backwards compatibility.

Would it be possible to put these tests outside of the `pre13` directory?

mottosso:
Just so I'm on the same page, what triggered the need for this functionality from you? What did you try to do, where you noticed this was a problem?

If you could post a usage example of what you couldn't do before, but could do with this feature, that would be most helpful!

p4vv37:
I've reverted the test from pre13 and moved this test.

I'm using this funcionality in my modifications to pyblish-lite for hiding plugins incompatible with checked instances:
https://github.com/p4vv37/pyblish-lite/tree/my_branch

I thought, that it would be wiser to use this function, instead of writing my own, to check if a plugin should be hidden.

mottosso:
> I thought, that it would be wiser to use this function, instead of writing my own, to check if a plugin should be hidden.

The thinking sounds right, thanks @p4vv37! I'll take a look at this as soon as I can.

pyblish/logic.py
```diff
@@ -153,10 +153,26 @@ def plugins_by_families(plugins, families):
 
     """
 
+    algorithms = {
```
mottosso: Since the same algorithms are being used in both this function, and `instances_by_plugin`, what do you think about moving this dictionary to the module-level, and prefix it with an underscore to highlight the fact that it is internal. That way, we could use it in both functions, and changing it would be simpler and more predictable.

mottosso (changes requested):
Great work on this, the tests look solid, style is consistent and functionality superb.

Have a look at the one note I made, and let me know what you think. This is nearly ready for merge and release.

Thanks again @p4vv37!

pyblish/logic.py
p4vv37: Right, that's a nice idea. I'm committing the change.

mottosso (approved):

mottosso:
This looks good to me. Happy for me to merge this?

p4vv37:
Sure, go ahead

PR #312, opened 2017-06-02
CLI Gui

tokejepsen:
This is my suggestion for launching a gui with the terminal. This is not ready for merging, as pyblish-qml isn't fully functioning.

Register pyblish_lite and show gui:
```
$ pyblish --register-gui pyblish_lite publish --gui
```
You can also register a host:
```
$ pyblish --register-gui pyblish_lite --register-host myHost publish --gui
```

Current issue is that when launching pyblish_qml the UI appears and disappears straight way.

### DIFF
--- a/pyblish/cli.py
+++ b/pyblish/cli.py
@@ -20,9 +20,11 @@ import os
 import time
 import json
 import logging
+import sys
 
 from . import api, lib, util, __version__
 from .vendor import click
+from .vendor.Qt import QtWidgets
 
 _ctx = None
 _help = {
@@ -42,7 +44,10 @@ _help = {
         "version": "Print the current version of Pyblish",
         "plugins": "List all available plugins",
         "data": "Initialise context with data. This takes "
-                "two arguments, key and value."
+                "two arguments, key and value.",
+        "register-gui": "Register a gui with its module name to use. For "
+                        "example register pyblish-qml with \"pyblish_qml\".",
+        "register-host": "Register a host"
     },
     "publish": {
         "delay": "Add an artificial delay to each plugin. "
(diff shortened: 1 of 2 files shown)


mottosso:
Some notes.

1. We shouldn't need to embed Qt.py, as (1) pyblish-lite embeds this itself and pyblish-qml doesn't support it, (2) I'd like it to support GUIs not necessarily written with Qt.
2. As we now have quite a few things to register, I'm thinking we may be able to concatenate the various commands into a single command. That way, it'd reduce the number of individual registry functions and as an added bonus also resemble the current syntax for adding data, as key/value pairs. See example below.

**Example below**

```bash
$ pyblish --register gui pyblish_lite --register host MyHost
```

For showing the GUIs, they both offer an interface for independent use via `python -m pyblish_lite` and `python -m pyblish_qml`. But those make it tricky to pass context and other data to them, so it's either we (1) find a way for them to pick up on data and hosts etc. through the environment, in which case we can insert these things there and have both GUIs read up on it or (2) add a similar interface to each gui, i.e. `--data` and `--register` etc. Though in that case we might as well skip implementing those things here and resort to simply using the individual GUIs, making sure they offer an identical interface throughout.

I'm thinking that what we should do is add the Pyblish CLI to each GUI.

```bash
$ python -m pyblish_qml --data key value --register host MyHost
```

It won't require `click` as Pyblish does, as it's a straightforward interface where `argparse` should suffice.

For dynamically picking a GUI in a host such as AE, we should then be able to get away with something like..

```bash
$ set GUI=pyblish_lite
$ python -m %GUI% --data key value --register host MyHost
```

tokejepsen:
This has been superseeded by https://github.com/pyblish/pyblish-base/pull/313

PR #313, opened 2017-06-02
Running a GUI from the CLI

mottosso:
Here's another method of running any GUI from the CLI. Data being passed in through a transient plug-in, generated at run-time.

```bash
$ python -m pyblish --data key value gui pyblish_qml
```


tokejepsen:
That is interesting :) I'm game for it, although seems slightly like a work around.

We would just need to have a "register_host" command flag.

mottosso:
> although seems slightly like a work around.

It's a little hacky, but it'll work on any GUI, both present and future, won't need individual maintenance as we're building off of regular plug-ins and is extensible in the sense that we could embed any kind of plug-in here, including those that fetch data dynamically like from a DB.

tokejepsen:
When running the command I get a key word error, did you test succesfully on your end?
```
$ python -m pyblish --data key value gui pyblish_qml
  File "C:\Users\tje\Documents\conda-git-deployment\windows\miniconda\envs\tgbvfx-environment\lib\runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\Users\tje\Documents\conda-git-deployment\windows\miniconda\envs\tgbvfx-environment\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "c:\users\tje\documents\conda-git-deployment\repositories\tgbvfx-environment\pyblish-base\pyblish\__main__.py", line 10, in <module>
    cli.main(obj={}, prog_name="pyblish")
  File "c:\users\tje\documents\conda-git-deployment\repositories\tgbvfx-environment\pyblish-base\pyblish\vendor\click\core.py", line 488, in __call__
    return self.main(*args, **kwargs)
  File "c:\users\tje\documents\conda-git-deployment\repositories\tgbvfx-environment\pyblish-base\pyblish\vendor\click\core.py", line 474, in main
    self.invoke(ctx)
  File "c:\users\tje\documents\conda-git-deployment\repositories\tgbvfx-environment\pyblish-base\pyblish\vendor\click\core.py", line 758, in invoke
    return self.invoke_subcommand(ctx, cmd, cmd_name, ctx.args[1:])
  File "c:\users\tje\documents\conda-git-deployment\repositories\tgbvfx-environment\pyblish-base\pyblish\vendor\click\core.py", line 767, in invoke_subcommand
    return cmd.invoke(cmd_ctx)
  File "c:\users\tje\documents\conda-git-deployment\repositories\tgbvfx-environment\pyblish-base\pyblish\vendor\click\core.py", line 659, in invoke
    ctx.invoke(self.callback, **ctx.params)
  File "c:\users\tje\documents\conda-git-deployment\repositories\tgbvfx-environment\pyblish-base\pyblish\vendor\click\core.py", line 325, in invoke
    return callback(*args, **kwargs)
TypeError: gui() got an unexpected keyword argument 'gui'
```

mottosso:
No, you're right. I was really rushing this PR out. Give it a try now.

tokejepsen:
That works for me :)
Tested pyblish_lite as well. What do we need to merge this?

mottosso:
> What do we need to merge this?

Some tests, and your `register_host` should do. Could you take over from here? Otherwise it might take a while before I can get back on it.

You can push directly to my fork/branch and it'll update this PR.

tokejepsen:
> You can push directly to my fork/branch and it'll update this PR.

Sure, just need to clean up the other PRs.

mottosso:
For `--register-host`, I think we'll be best off storing the hosts in an environment variable, but I'm not sure there is one for hosts yet.

If we store it in the environment, it'll be that much easier to pass to GUIs, without explicitly implementing support for this in each GUI.

mottosso:
For example,

```python
if hosts:
  environ["PYBLISH_HOSTS"] = os.pathsep.join(hosts)
...
subprocess.call(...)
```

tokejepsen:
> If we store it in the environment, it'll be that much easier to pass to GUIs, without explicitly implementing support for this in each GUI.

Do you think that should be in the core of Pyblish, rather than the CLI?

mottosso:
> Do you think that should be in the core of Pyblish, rather than the CLI?

Yes, we'll need to make Pyblish consider hosts from `register_host()` and hosts from the environment. `register_gui()` and `register_plugin_path()` should provide some guidance on how it can be done. It'd be fine to do in the same PR, as they are related.

tokejepsen:
The environment variable does look a bit odd with double "H", but ```PYBLISHPLUGINPATH``` is without underscores so copying that. What do you think?

Where should the guidance about the environment variables live? https://api.pyblish.com/ ?

Lastly what kind of tests should we be doing? Very unfamiliar with tests, so don't quite know where to start.

mottosso:
> What do you think?

I think we should introduce an underscore. I think this should happen across all environment variables, they are a little un-uniform at the moment. This could be a start of that.

> Where should the guidance about the environment variables live?

Yes, in the API docs should be good.

> Lastly what kind of tests should we be doing?

Have a look at how the CLI is tested currently, it's essentially running commands like from the command line, and asserting the textual output. You can also gain access to the resulting context and inspect this too.

tests/test_api.py
```diff
@@ -0,0 +1,28 @@
+import os
+
+import pyblish.api
+
+
+def test_environment_host_registration():
```
mottosso: Thanks for this.

I'm not sure this is the right approach. Maybe a better test would be to include a plug-in that you've associated to a host, and test for whether it runs as expected? That way you could test it both via `util.publish` and the CLI equivalent.

tokejepsen: How about this?

tests/test_logic.py
```diff
@@ -69,7 +69,7 @@ def process(self, instance):
 
 def test_register_gui():
     """Registering at run-time takes precedence over those from environment"""
-    
+
     with no_guis():
         os.environ["PYBLISHGUI"] = "second,third"
```
tokejepsen, line 74: I have never actually used ```PYBLISHGUI```. When was this introduced?
Should this be ```PYBLISH_GUI```instead?
Also finding is odd to use comma instead of os path separator, when working with environment variables.

mottosso: > I have never actually used PYBLISHGUI. When was this introduced?

If you browse to it in the code, and click the `Blame` button, it'll show you when.

> Should this be PYBLISH_GUI instead?

Yes, indeed it should. If you jump on this, don't forget to keep the old one in place as well, for backwards compatibility. We'll document the new one, and then one day in the future we can deprecated it.

tokejepsen:
I would like to test whether a GUI actually shows or not. Since this means that you need a gui's module present in the testing environment, I'm guessing this it not very desirable?

mottosso:
> I'm guessing this it not very desirable?

That's a good idea. You can create a mock gui, a module with a `show()` function that you register and check whether it got called or not.

tokejepsen:
Ahh, ok. I'm guessing CLI testing of the individual GUIs should happen in their repositories.

tests/test_logic.py
tokejepsen: Updated with PYBLISH_GUI support.

tokejepsen:
In order to test showing the GUI with a mock up, I had to change how the GUI's are executed with ```subprocess```, from ```subproces.call``` to ```subprocess.check_output```.
One problem with this is that we don't get a streamed output from the GUI, instead we get the whole output of when exiting the GUI.

tokejepsen:
I'm getting bad exit codes on python3 and python26. I'm developing on python27, which passes the exit test. Do you have any idea why python3 and python26 is failing?

mottosso:
Hard to tell, you'll need to read from the `stdout` of the failed process, and see what exception it threw.

tokejepsen:
Wow, took a few tries to pass those tests :)

I think I have covered the introduced code with tests, but let me know if you have any more ideas.

mottosso:
Thanks, having a look in a bit. :)

pyblish/cli.py
```diff
@@ -303,4 +334,42 @@ def publish(ctx,
...
+
+    with _cli_plugin(data=context.data) as plugin_path:
+        environ["PYBLISHPLUGINPATH"] = os.pathsep.join(
+            ctx.obj["plugin_paths"] + [plugin_path]
+        )
+
+        process = subprocess.Popen(
+            [sys.executable, "-m", package + ".__main__"],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
```
mottosso: Tip: If you make this `subprocess.STDOUT` instead, errors will be passed through stdout, and you'll only need to look for output there.

pyblish/cli.py
```diff
@@ -303,4 +334,42 @@ def publish(ctx,
...
+            env=environ,
+            bufsize=1,
+            universal_newlines=True
+        )
+        while True:
```
mottosso, line 358: Listening to `stdout` is tricky. All sorts of things can be printed from the host, and some things differ across Python 2 and 3.

I can't say for sure whether the approach you have here will *not* work without issues, but I *can* say that the below example *would* work. If you don't mind, I'd like to see this in there instead.

```python
for line in iter(process.stdout.readline, b""):
  ...
```

- [Reference](https://github.com/pyblish/pyblish-qml/blob/84cc26c560b8a8a96d57fec06104971f085d3c57/pyblish_qml/ipc/server.py#L176)

pyblish/logic.py
```diff
@@ -134,9 +134,14 @@ def show():
 
 def registered_guis():
     """Return registered GUIs"""
-    from_environment = os.environ.get("PYBLISHGUI", "")
+    # Backwards compatibility for PYBLISHGUI
+    old_environment = os.environ.get("PYBLISHGUI", "")
```
mottosso: I think it's safe to make these either-or.

```python
# E.g.
from_environment = os.environ.get("PYBLISH_GUI", os.environ("PYBLISHGUI", ""))
```

tests/mock_gui/app.py
```diff
@@ -0,0 +1,13 @@
+import pyblish.cli
```
mottosso: I'd rather we make these written from within the test than hosted as individual files.

I've been bitten by this once before, the problem arises when we update Pyblish and this interface. We'll still want to test for the old behaviour, for backwards compatibility, but also test for the new one. So then you'll need to duplicate these files.

See the [`preXX`](https://github.com/pyblish/pyblish-base/tree/master/tests) folders I had to do to combat this..

If you write them into a temporary folder in the `setup()` of the test module, it'll run once before all tests. Then in `teardown()` you can `shutil.rmtree` it.

mottosso:
Looks good! Just a few minor notes, let me know what you think.

pyblish/cli.py
tokejepsen: Cool, thanks :)

pyblish/cli.py
tokejepsen: On both Python2.7 and Python3.5 changing this, means it'll no longer spit out any error messages.

tokejepsen: Also we miss out on some messages from stdout in for example pyblish-qml:
With ```for line in iter(process.stdout.readline, b""):```:
```
Using Python @ 'C:\Users\tje\Documents\conda-git-deployment\windows\miniconda\envs\temp27\python.exe'
Starting pyblish-qml
Entering state: "hidden"
Entering state: "clean"
Settings:
  HeartbeatInterval = 60
  WindowSize = [430, 600]
Entering state: "visible"
Spent 83.00 ms resetting
Spent 282.00 ms resetting
Entering state: "ready"
```
Without ```for line in iter(process.stdout.readline, b""):```:
```
Starting pyblish-qml server..
Using Python @ 'C:\Users\tje\Documents\conda-git-deployment\windows\miniconda\envs\temp27\python.exe'
Using PyQt5 @ 'None'
Starting pyblish-qml
Done, don't forget to call `show()`
Entering state: "hidden"
Entering state: "ready"
Entering state: "clean"
Entering state: "alive"
Settings:
  WindowTitle = Pyblish
  HeartbeatInterval = 60
  WindowPosition = [100, 100]
  WindowSize = [430, 600]
  ContextLabel = Context
Entering state: "visible"
Entering state: "initialising"
Spent 82.00 ms resetting
Made 0 requests during publish.
Spent 213.00 ms resetting
Made 0 requests during reset.
Entering state: "ready"
No local comment, reading from context..
```

mottosso: > On both Python2.7 and Python3.5 changing this, means it'll no longer spit out any error messages.

Was this after you changed to `STDOUT` above? This loop would only listen on stdout.

tokejepsen: Yeah, changed the subprocess argument to ```stderr=subprocess.STDOUT```, and not getting any error messages.

pyblish/logic.py
tokejepsen: Done :)

pyblish/cli.py
mottosso: Most interesting.. let's keep it the way you've got it in that case, I'll keep an eye out.

tests/mock_gui/app.py
tokejepsen: Didn't know exact where you wanted this to happen, so I have made a context within the test module for creating the mock gui.
Do you think it should be per test instead of reusing it across multiple tests?

tests/test_cli.py
```diff
@@ -20,6 +25,51 @@ def context():
     return ctx().obj["context"]
 
 
+mock_gui_main = """from .app import show
```
mottosso: This could all be just a single callable module, no need for a package.

tokejepsen: The problem is that in order to test how pyblish-base is calling gui, ei. ```sys.executable, "-m", package + ".__main__"```, we need it to be a package.
Or maybe I'm missing something?

tests/test_cli.py
```diff
@@ -67,3 +117,92 @@ def process(self, instance):
...
+    print(result.output)
+
+    assert count["#"] == 11, count
+
+
+@with_setup(lib.setup, lib.teardown)
+def test_show_gui():
+    """Showing GUI through cli works"""
+
+    with mock_gui() as path:
```
mottosso: I think this is a good approach.

tests/test_cli.py
tokejepsen: @mottosso did you have another method for this?

mottosso: You can call a module as-is.

You can't call a package as-is, which is where `__main__.py` comes in. It makes a package executable.

Simply move everything into a single file, call it `sys.executable, "-m", "mock_gui"` and it should work.

tokejepsen: The problem is that you can't do ```sys.executable, "-m", "mock_gui"``` on python 2.6 https://stackoverflow.com/a/6630931

mottosso: I think he's referring to a Python *package* called `mock_gui`, and not a module (file). Are you sure you've tried calling just the module?

tokejepsen: No, I didn't actually test this with ```mock_gui```.

I tested with ```pyblish-lite``` on python 2.6.

```
(temp26) C:\Users\tje\Documents\conda-git-deployment\repositories\tgbvfx-environment\pyblish-base>python -m pyblish_lite
C:\Users\tje\Documents\conda-git-deployment\windows\miniconda\envs\temp26\python.exe: pyblish_lite is a package and cannot be directly executed
```

tokejepsen: @mottosso how do you think we should approach this?

mottosso: I had a go with writing these tests. I'm not sure what the previous tests were doing, the Mock GUI was importing Click? That's some inception stuff going on there.

Have a look at what you think of the current tests.

mottosso: So we're on the same page, what the `gui` argument is meant to do is call on the module you give it, and assumes it will run something - in most cases, show the GUI. For testing purposes, it's enough to simply test for whether it runs or not.

For data, the important bit is that the resulting call to the module you give it is given access to it. From an actual GUI, the data would be available to the plug-ins that run within that GUI. We can test that by simply calling `util.publish()` and inspecting the resulting `context`.

tokejepsen:
This all looks good to me, just need to get rid of the "tests/mock_gui" redundant module files.

mottosso:
This looks good to me now!

tokejepsen:
Excellent! When this is merged we can fully go back to pyblish-qml :)

mottosso:
Let's do it!

PR #314, opened 2017-06-08
fixed issue with plugins on windows platform

SegaMegaDrive16:
Plugins could`t be registered by register_plugins_path method

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -1265,7 +1265,7 @@ def discover(type=None, regex=None, paths=None):
 
             try:
                 with open(abspath) as f:
-                    six.exec_(f.read(), module.__dict__)
+                    six.exec_(f.read().replace('\r', ''), module.__dict__)
 
                 # Store reference to original module, to avoid
                 # garbage collection from collecting it's global


mottosso:
Thanks @GreenGoseWithFunnyHat, are you able to provide an example I could use to reproduce this problem? I'll use that to create a test for it, so the problem is automatically caught the next time it appears.

SegaMegaDrive16:
I guess that windows plaform (tested on widnows 7 with different file editors) creates \r\n instead of \n on linux at the end of each string. ( I have found [discussion](https://superuser.com/questions/374028/how-are-n-and-r-handled-differently-on-linux-and-windows) about this, they say that ` MS-DOS chose CR+LF, and Windows inherited this`.
To catch this problem you need to test this on windows 7 platfrom ( i guess in 10th version it was fixed ) or replace '\n' to '\r\n' in you plugins file.
If you need i can attach file with \r\n at the end of strings.

mottosso:
Would it be possible to create a small sample, and upload it here? That way I can include this in the automated test suite.

mottosso:
> If you need i can attach file with \r\n at the end of strings.

Yes, please. :)

BigRoy:
@GreenGoseWithFunnyHat have you been able to find some time to set up a reproducable example and maybe also implement a test that ensures it's safe with these commits?

Issue #306, opened 2016-11-18
Module mixup with same named files.


tokejepsen:
**Description**

If you have files named the same, the associated imported modules gets mixed up.

**Expected results**

Would expect to be able to have same named files, and get the correct imports.

**Short reproducible**

To reproduce we need three files:
- publish.py (file)
    - first_plugin (directory)
        - collect.py (file)
    - second_plugin (directory)
        - collect.py (file)

```publish.py```
```python
import os

import pyblish.util


os.environ["PYBLISHPLUGINPATH"] = (os.path.join(os.path.dirname(__file__),
                                                "first_plugin") +
                                   os.pathsep +
                                   os.path.join(os.path.dirname(__file__),
                                                "second_plugin"))

context = pyblish.util.publish()
for result in context.data["results"]:
    for record in result["records"]:
        print record.getMessage()
    if result["error"]:
        print result["error"]
```
```first_plugin/collect.py```
```python
import os

import pyblish.api


class FirstCollect(pyblish.api.ContextPlugin):

    order = pyblish.api.CollectorOrder

    def process(self, context):

        self.log.info("Hello from first plugin!")
        self.log.info(os.path)
```
```second_plugin/collect.py```
```python
import pyblish.api


class SecondCollect(pyblish.api.ContextPlugin):

    order = pyblish.api.CollectorOrder

    def process(self, context):

        self.log.info("Hello from second plugin!")
```

When you execute ```publish.py``` Pyblish will execute ```second_plugin/collect.py``` then ```first_plugin/collect.py```. I think since Pyblish have registered the ```collect``` module first with ```second_plugin/collect.py```, that is also the import modules it'll get when executing ```first_plugin/collect.py```, hence the missing ```os``` module.

mottosso:
Sorry, what was the error?

tokejepsen:
Good point. Here is the expected output from the reproducible:

```
Hello from second plugin!
Hello from first plugin!
'NoneType' object has no attribute 'path'
```

mottosso:
Ah I see.

Yes this is a tricky one. The reason is due to the fact that plug-ins are "imported" anew on each discovery. The standard Python mechanism would require you to reload the module explicitly, or restart the interpreter. But we don't want that, because it would prohobit efficient development of plug-ins in hosts, such as Maya.

A way to work around it, is by putting your imports in your `process()` method.

``` python
def process(self, context):
  import os
  self.log.info(os.path)
```

[...]

PR #317, opened 2017-06-19
Initial solution proposal for fixing #306

tokejepsen:
#306

I need a bit of help on this :(

I've replicated the test from the issue, and two others, along with the solution. The tests pass here, but when I'm NukeStudio I still get the same problem with imports.

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -1260,6 +1260,17 @@ def discover(type=None, regex=None, paths=None):
             if not mod_ext == ".py":
                 continue
 
+            def replace_non_ascii(unicode_string):
+                result = u""
+                for i in range(0, len(unicode_string)):
+                    if ord(unicode_string[i]) < 128:
+                        result += unicode_string[i]
+                    else:
+                        result += "?"
+                return str(result)
+
+            mod_name = replace_non_ascii(os.path.join(path, mod_name))
+
             module = types.ModuleType(mod_name)
             module.__file__ = abspath
 
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -733,3 +733,110 @@ def test_changes_to_registered_plugins_are_not_persistent():
 
     registered = pyblish.api.registered_plugins()[0]
     assert registered.active is False
+
+
+@with_setup(lib.setup_empty, lib.teardown)
+def test_same_named_plugins():
+    """Having same named plugins works"""
+
+    with lib.tempdir() as temp:
+        first_module = os.path.join(temp, "first_plugin", "collect.py")
+        print("Writing temporarily to: %s" % first_module)
+        pyblish.api.register_plugin_path(os.path.dirname(first_module))
+
+        if not os.path.exists(os.path.dirname(first_module)):
+            os.makedirs(os.path.dirname(first_module))
+
+        with open(first_module, "w") as f:
+            f.write("""
+import os
+import pyblish.api
+
+class FirstCollect(pyblish.api.ContextPlugin):
+
+    order = pyblish.api.CollectorOrder
+
+    def process(self, context):
+        self.log.info(os)
+""")
+
+        second_module = os.path.join(temp, "second_plugin", "collect.py")
+        print("Writing temporarily to: %s" % second_module)
+        pyblish.api.register_plugin_path(os.path.dirname(second_module))
+
+        if not os.path.exists(os.path.dirname(second_module)):
+            os.makedirs(os.path.dirname(second_module))
+
+        with open(second_module, "w") as f:
+            f.write("""
+import pyblish.api
+
+class SecondCollect(pyblish.api.ContextPlugin):
+
+    order = pyblish.api.CollectorOrder
+
+    def process(self, context):
+        pass
+""")
+
+        context = pyblish.util.publish()
+
+        log_message = None
+        for result in context.data["results"]:
+            if result["records"]:
+                log_message = result["records"][0].message
+
+        assert log_message == str(os)
+
+
+@with_setup(lib.setup_empty, lib.teardown)
+def test_namespace_is_path():
+    """The namespace of the module is the same as the path"""
+
+    with lib.tempdir() as temp:
+        module = os.path.join(temp, u"temp")
+        module_file = module + ".py"
+        print("Writing temporarily to: %s" % module_file)
+        pyblish.api.register_plugin_path(temp)
+
+        with open(module_file, "w") as f:
+            f.write("""
+import pyblish.api
+
+class Collect(pyblish.api.ContextPlugin):
+
+    def process(self, context):
+        pass
+""")
+
+        plugins = [p.__module__ for p in pyblish.api.discover()]
+        assert module in plugins, plugins
+
+
+@with_setup(lib.setup_empty, lib.teardown)
+def test_non_ascii_plugin_name():
+    """Having non-ascii plugin name works"""
+
+    with lib.tempdir() as temp:
+        module = os.path.join(temp, u"M\u00E5rten.py")
+        print("Writing temporarily to: %s" % module)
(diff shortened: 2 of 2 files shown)


mottosso:
Just so I'm on the same page about your problem.

You have two files of the same name, in different directories. You would like to register both directories, and read the plug-ins from each of them.

```python
from pyblish import api
api.register_plugin_path("c:/path/1")
api.register_plugin_path("c:/path/2")
```

But you can't give them different names.

**E.g.**

```
first_plugin/
  first_collect.py
second_plugin/
  second_collect.py
```

Why?

I think the reason I haven't been looking at this as a real problem is because I don't understand what the problem is in giving each of them unique names. If you could help me understand how this isn't about cosmetics of files on disk, and really does prevent you from accomplishing a particular task then it'd be easier for me to think about this critically.

tokejepsen:
> I think the reason I haven't been looking at this as a real problem is because I don't understand what the problem is in giving each of them unique names. If you could help me understand how this isn't about cosmetics of files on disk, and really does prevent you from accomplishing a particular task then it'd be easier for me to think about this critically.

After working on this, writing the tests and solution, I am somehow doubtful the problem is with same named plugins.
The original problem came when working on the pyblish-bumpybox set of plugins, where I got the import issues (see the results of the test: https://github.com/pyblish/pyblish-base/issues/306#issuecomment-261562349). It is these import issues that are preventing me some continuing. I honestly don't understand enough of pyblish to know where the problem is.

mottosso:
That error might be unrelated; it could have to do with the automatic reload feature of plug-ins.

Are you able to reproduce it now? If so, try adding all imports into the `process()` function.

mottosso:
Or wait, sorry, you already did post the reproducible.

tokejepsen:
> If so, try adding all imports into the process() function.

Adding all the external imports used in the process() function definitely works. So I could workaround the problem for now, but guessing we would want to resolve the issue instead.

> If you rename those modules, does the error still occur?

Renaming the modules in the reproducible works, which is why when I saw the same error in the pyblish-bumpybox plugins, I assumed it was the same issue. Though renaming the plugins in pyblish-bumpybox, more specific the Hiero/NukeStudio plugins, does not resolve the issue. Even tried other things like renaming the class and making the plugin exclusive to the "nukestudio" host.

mottosso:
> Adding all the external imports used in the process() function definitely works. So I could workaround the problem for now, but guessing we would want to resolve the issue instead.

Yes, that would be good I think, but that is a different issue. Once we've made paths unique, this will still be a problem.

It's because as soon as a module is reloaded, the old one is garbage collected and so all references to the old module from a newly reloaded one will be invalid.

Modules get garbage collected once there is no more references to it, and the only reference to modules is in the `sys.modules` dictionary.

You can test this for yourself by importing a module of your choice, and removing it.

```python
import sys
import my_module
sys.modules.pop("my_module")

my_module.my_function()
# ERROR
```

Now because each plug-in is a class, and the class references things from the module-level, such as your module-level imports, the class will be trying to reach garbage collected global references. This is why you are seeing these issues, and the reason it is resolved by putting all of your imports into the `process()` function; in there they are loaded and used in the same place with no room for one being garbage collected.

> Though renaming the plugins in pyblish-bumpybox, more specific the Hiero/NukeStudio plugins, does not resolve the issue. Even tried other things like renaming the class and making the plugin exclusive to the "nukestudio" host.

Ok, we should resolve this non-unique naming problem. But this sounds like a different issue we should resolve first.

Are you able to reproduce the error when each of your plug-in files have unique names?

tokejepsen:
> Are you able to reproduce the error when each of your plug-in files have unique names?

I checked and all the file names in pyblish-bumpybox are unique, so I'm drawing a blank for where the issue could be.

> Now because each plug-in is a class, and the class references things from the module-level, such as your module-level imports, the class will be trying to reach garbage collected global references. This is why you are seeing these issues, and the reason it is resolved by putting all of your imports into the process() function; in there they are loaded and used in the same place with no room for one being garbage collected.

Maybe we should store those references to the module level imports in sys.modules?

mottosso:
> I checked and all the file names in pyblish-bumpybox are unique, so I'm drawing a blank for where the issue could be.

Could you help me reproduce this?

tokejepsen:
Save this to ```check_unique_names.py```:
```python
import os


plugins = {}
for root, dirs, files in os.walk("."):
    for f in files:
        mod_name, ext = os.path.splitext(f)

        if not f.endswith(".py"):
            continue

        path = os.path.join(root, f)
        if f in plugins.keys():
            plugins[mod_name].append(path)
        else:
            plugins[mod_name] = [path]

for key, value in plugins.iteritems():
    if len(value) >= 2:
        print value
```
In terminal:
```
cd [pyblish-bumpybox repository]
python /path/to/check_unique_names.py
```

tokejepsen:
This test shows the error I'm getting in NukeStudio, basically loosing the reference to the module imports:
```python
@with_setup(lib.setup_empty, lib.teardown)
def test_same_named_plugins():
    """Having same named plugins works"""

    with lib.tempdir() as temp:
        first_module = os.path.join(temp, "first_plugin", "collect.py")
        print("Writing temporarily to: %s" % first_module)
        pyblish.api.register_plugin_path(os.path.dirname(first_module))

        if not os.path.exists(os.path.dirname(first_module)):
            os.makedirs(os.path.dirname(first_module))

        with open(first_module, "w") as f:
            f.write("""
import os
import pyblish.api

class FirstCollect(pyblish.api.ContextPlugin):

    def process(self, context):
        self.log.info(os)
""")

        second_module = os.path.join(temp, "second_plugin", "collect.py")
        print("Writing temporarily to: %s" % second_module)
        pyblish.api.register_plugin_path(os.path.dirname(second_module))

        if not os.path.exists(os.path.dirname(second_module)):
            os.makedirs(os.path.dirname(second_module))

        with open(second_module, "w") as f:
            f.write("""
import pyblish.api

class SecondCollect(pyblish.api.ContextPlugin):

    def process(self, context):
        pass
""")

        log_message = None
        for result in context.data["results"]:
            if result["records"]:
                log_message = result["records"][0].message

        assert log_message == str(os)
```

mottosso:
Isn't that using identical names?

I mean, can you help me reproduce a situation where the files have unique names and still throw an error?

tokejepsen:
> I mean, can you help me reproduce a situation where the files have unique names and still throw an error?

Sorry no, it seems to only be in NukeStudio, and I can't reproduce outside cause I have no idea about where the problem is.

tokejepsen:
Although it throws the same error as the repoducible, so its loosing the reference to the imported modules.

mottosso:
Ok. Does the solution in this PR address the issue?

tokejepsen:
No, it does not resolve the issue unfortunately. Going to close this and resort to having import statements in the class.

mottosso:
Ok, for completeness and when we return to this topic, there are two separate but similar issues at play here.

1. The files within which plug-ins reside must have unique names, because their names are used as identifiers in `sys.modules["plugin_file_here"]` and any identical name loaded after another would overwrite the first one.
2. Automatic reload breaks module-level references.

The reason this PR didn't resolve (2) was because the similarities of names wasn't the issue, automatic reload was.

Solving that would involve looking at the `discover()` function and either (1) disable automatic reload or (2) somehow "merging" the differences between a newly loaded module and the one loaded before it. At the moment, it's doing a plain overwrite, without regards to what was there before, and it's possible we could take inspiration from Git and try to establish (1) which attributes and methods on each contained plug-in class changed and integrate these changes more delicately.

The problem there is that if we do it not-so-elegantly, we run into the risk of obscure run-time errors, the kind that goes away when you "restart Maya".

There are other plug-in frameworks out there for Python, it's possible they have thought about this harder and come up with a solution. But what most of them do, including Maya's plug-in mechanism, is simply not do automatic reload. That would simplify things greatly.

PR #318, opened 2017-06-20
Fix Module mixup with same named files (#306)

### LINKED ISSUE #306: Module mixup with same named files.

mottosso:
This alters the name of plug-ins used in-memory to one that is *almost* guaranteed to be unique.

Edge cases include directories with similar characters in similar places.

```bash
c:\my\plögins\collect.py
c:\my\plågins\collect.py
```

Though different, would both result in..

```bash
c__my_pl_gins_collect_py
c__my_pl_gins_collect_py
```

Where the latter one overwrites the former.

### DIFF
--- a/pyblish/api.py
+++ b/pyblish/api.py
@@ -157,6 +157,7 @@ def __init__():
     # Register default test
     register_test(__default_test)
 
+
 __init__()
 
 
@@ -165,6 +166,11 @@ __all__ = [
     "Context",
     "Instance",
     "Asset",
+
+    # Matching algorithms
+    "Subset",
+    "Intersection",
+    "Exact",
 
     "Plugin",
     "Action",
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -1260,11 +1260,7 @@ def discover(type=None, regex=None, paths=None):
             if not mod_ext == ".py":
                 continue
 
-            # Remove all characters that isn't in a-z, A-Z and 0-9
-            # See #306 for details
-            safe_name = re.sub(r'([^\w])', "_", abspath)
-
-            module = types.ModuleType(safe_name)
+            module = types.ModuleType(mod_name)
             module.__file__ = abspath
 
             try:
@@ -1274,7 +1270,7 @@ def discover(type=None, regex=None, paths=None):
                 # Store reference to original module, to avoid
                 # garbage collection from collecting it's global
                 # imports, such as `import os`.
-                sys.modules[safe_name] = module
+                sys.modules[abspath] = module
 
             except Exception as err:
                 log.debug("Skipped: \"%s\" (%s)", mod_name, err)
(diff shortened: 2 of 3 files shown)


tokejepsen:
This seems to fix an issue I'm having currently in Nuke.

Should we give this a merge?

mottosso:
I've been experimenting with a different approach lately, give this a try and let me know if that works for you.

tokejepsen:
Unfortunately this does not fix the issue I had in Nuke now :(

mottosso:
Ok, let me back it up.

mottosso:
Ok, should be back to normal, with the exception that it merges with master and the target feature.

tokejepsen:
Yup, we are back, and the issue in Nuke is fixed :)

mottosso:
Before merging this, would you mind having a look at the failing test? Otherwise I'll try and have a look at it when I can.

tokejepsen:
The problem with the test is that its trying to get the module of the plugin through the module name in ```sys.modules``` but since we have changed that to the absolute path, it can't find the module anymore.

Also I noticed that you are not replace ```\\``` with ```_```, which you outlined in the PR. Not a problem for me or Nuke :)

tokejepsen:
Question is how to access the absolute path to the module from the class?

tokejepsen:
Latest commit seems to fix the issue :)

mottosso:
Yeah, that should work fine. Thanks, let's give this a try. :)

PR #319, opened 2017-08-16
Test for logging only pyblish records.

tokejepsen:
This is an effort to get pyblish to only record log records related to Pyblish. This means log calls with ```self.log```.

### DIFF
--- a/pyblish/lib.py
+++ b/pyblish/lib.py
@@ -49,7 +49,8 @@ class MessageHandler(logging.Handler):
         self.records = records
 
     def emit(self, record):
-        self.records.append(record)
+        if record.name.startswith("pyblish"):
+            self.records.append(record)
 
 
 def extract_traceback(exception):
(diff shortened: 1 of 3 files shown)


tokejepsen:
This is spawned from a discussion about non-ascii characters coming from other loggers; https://github.com/pyblish/pyblish-qml/issues/251

mottosso:
Indeed, that catches the nail on the head. If we can make this test pass, it should resolve your ftrack issue as well.

tokejepsen:
What about this solution?

Tests are passing, and I've also tested with Ftrack and it solves the issue there as well.

mottosso:
That's.. ingenious. If you could let this run for a few days in production, I see no issue with it. Good job.

tokejepsen:
This has been working flawlessly :)

mottosso:
Good. :)

Patch-version up and merge when ready!

tokejepsen:
Ready for a merge :) Can't merge myself.

mottosso:
Thanks @tokejepsen!

tokejepsen:
Guess it needs a release as well?

mottosso:
All done!

https://github.com/pyblish/pyblish-base/releases/tag/1.4.6

PR #320, opened 2017-08-18
Targets Workflow

tokejepsen:
**Motivation**

This PR is to enable a workflow where you can define targets to aim you plugins for.

The targets work the same as the family, but will eventually be implemented in the GUIs differently.


tokejepsen:
@mottosso What do you think about this solution?

Targets works in a similar way to families, but instead of comparing instances to a plugins targets, the plugins are filtered by the registered targets ```pyblish.api.registered_targets```.

Since all plugins are registered with ```*``` as targets, this workflow is backwards compatible with existing plugins that has not overwritten the ```targets``` attribute.

Expecting ```pyblish-qml``` and ```pyblish-lite``` to implement a method for registering targets per session, so you can do this:
```python
import pyblish_qml

pyblish_qml.show(targets=["custom_target"])
```

pyblish/logic.py
```diff
@@ -236,6 +238,37 @@ def plugins_by_host(plugins, host):
     return compatible
 
 
+def plugins_by_targets(plugins):
```
mottosso: Let's pass in targets ourselves, so it resembles the other `by_something()` functions.

`def plugins_by_targets(plugins, targets)`

mottosso:
> Since all plugins are registered with * as targets, this workflow is backwards compatible with existing plugins that has not overwritten the targets attribute.

Goooood. That is critically important, thanks for testing and pointing this out.

Very well done. Just that one nitpick, then I'm happy to merge.

tokejepsen:
All ready for a merge :)

pyblish/logic.py
```diff
@@ -236,6 +238,37 @@ def plugins_by_host(plugins, host):
     return compatible
 
 
+def plugins_by_targets(plugins, targets):
+    """Reutrn compatible plugins `plugins` to targets `targets`
+
+    Arguments:
+        plugins (list): List of plugins
+        families (list): Families with which to compare against
```
mottosso: Don't forget about this little fella.

tokejepsen: Good spot!

mottosso:
Looks good to me.

tokejepsen:
Ready for merge and release :)

mottosso:
Would you be kind enough to make a release statement for this one? Ideally with an example or two, along with motivation and usecase.

I've added you to the contributors list, where I thought you already were but there you go. :)

tokejepsen:
> Would you be kind enough to make a release statement for this one?

Sure, but I think I have done the versioning up incorrectly. Its currently ```1.5.6```, but it should be ```1.5.0```, right?

mottosso:
Ah, yes that's true. You can fudge the number without a PR for this one, just hit the edit button on GitHub and [make that change](https://www.youtube.com/watch?v=MLZ-ptqSjas).

PR #321, opened 2017-08-22
Revise default target

tokejepsen:
**Motivation**

The default target for plugins of ```*``` was confusing when dealing with logic that exclusively targeted plugins.
Further the ```plugins_by_targets``` has been revised to not assume a wildcard target to facilitate using the method outside of pyblish-base. This method is now exposed in the ```pyblish.api``` as well.
For these reasons the default target has been changed to ```default```.

### DIFF
--- a/pyblish/api.py
+++ b/pyblish/api.py
@@ -98,6 +98,7 @@ from .logic import (
     plugins_by_family,
     plugins_by_host,
     plugins_by_instance,
+    plugins_by_targets,
     instances_by_plugin,
     register_test,
     deregister_test,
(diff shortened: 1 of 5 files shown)


tokejepsen:
Very odd that I'm not getting the error locally, that Appveyor and Travis flags.

tokejepsen:
Turns out that the ```lib.setup_empty``` was not deregistering the targets.

tokejepsen:
Have written tests for targets and the matching algorithm to ensure future compatibility.

tokejepsen:
This has been used for nearly a week, and worked rather well. Merging.

PR #322, opened 2017-08-23
Publish with targets.

tokejepsen:
**Motivation**

Publish to targets without a GUI.

The use case is for a quick submission to a render farm without having to fiddling or wait for GUI.


tokejepsen:
The errors on Appveyor and Travis are because ```lib.setup_empty``` is not deregistering the targets.
The fix for that is coming in #321

tokejepsen:
Tests are passing :) Merging.

PR #323, opened 2017-09-04
Explicit target default plugins.

tokejepsen:
**Motivation**

Only include default plugins when explicitly requesting so.

The problem case is when you have default targeted collectors that produce instances, with plugins targeted to "default".

### DIFF
--- a/pyblish/logic.py
+++ b/pyblish/logic.py
@@ -354,12 +354,9 @@ def Iterator(plugins, context, state=None):
         "ordersWithError": set()
     }
 
-    targets = registered_targets()
-
     # We'll add "default" target if no targets are registered. This happens
     # when running the Iterator directly without registering any targets.
-    if not targets:
-        targets = ["default"]
+    targets = registered_targets() or ["default"]
 
     plugins = plugins_by_targets(plugins, targets)
 
--- a/pyblish/logic.py
+++ b/pyblish/logic.py
@@ -354,7 +354,13 @@ def Iterator(plugins, context, state=None):
         "ordersWithError": set()
     }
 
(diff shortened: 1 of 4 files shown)


tokejepsen:
@mottosso This PR is a maintenance release, with all tests passing. 

What is the workflow for merging? Merge when tests pass, or wait for other users input?

pyblish/util.py
```diff
@@ -12,7 +12,7 @@
 log = logging.getLogger("pyblish.util")
 
 
-def publish(context=None, plugins=None, targets=[]):
+def publish(context=None, plugins=None, targets=["default"]):
```
mottosso: Don't add it to the argument, in fact never put a list as default value for any argument; it's mutable. Meaning that if the function modifies the default value, it will persist till the next call to the function. It's a really tough quirk to debug. Better put `None` and make it a list in the function body.

pyblish/logic.py
```diff
@@ -354,7 +354,13 @@ def Iterator(plugins, context, state=None):
         "ordersWithError": set()
     }
 
-    targets = registered_targets() + ["default"]
+    targets = registered_targets()
+
+    # We'll add "default" target if no targets are registered. This happens
+    # when running the Iterator directly without registering any targets.
+    if not targets:
+        targets = ["default"]
```
mottosso: This could be written as `registered_targets() or ["default"]`

mottosso:
> What is the workflow for merging? Merge when tests pass, or wait for other users input?

I typically prefer leaving it for a day or two, give others a chance to chime in if they spot something. Now that time has passed and I'd be happy to merge given you address the minor notes.

PR #324, opened 2017-09-04
Validate publish data type

tokejepsen:
**Motivation**

When assigning any types other than boolean to an instance's data member ```publish``` it'll always return ```True```.

The data member have to be a boolean to work correctly.

**Implementation**

The solution is to add a built-in validator to validate all instances' ```publish``` data member.


mottosso:
Yeap, looks like it. If you delete the 2.6 entries from [here](https://github.com/pyblish/pyblish-base/blob/841f40821507964389adac01683ac86f745bb5fe/appveyor.yml#L3) it should pass. No need to test against it anymore.

tokejepsen:
Good to merge?

mottosso:
Please don't merge yet, I'm not sure about this functionality.

> When assigning any types other than boolean to an instance's data member publish it'll always return True.

What's wrong with that? Have you been seeing errors because of it (e.g. in Pyblish QML) or is it something logically confusing or..?

tokejepsen:
The issue was that an application (think it was Nuke) was returning a string rather than a boolean.
Without knowing this logic within the application I spend ages trying to figure out why when mapping an attribute to ```publish``` was always evaluating to ```True```.
This is purely to help someone else in the future not spend the same amount of time on this issue.

tokejepsen:
That could work :)

I've implemented it so the ```data``` member on both context and instance have this validation. Could split it up, but that would introduce maybe too much unnecessary code.

pyblish/plugin.py
```diff
@@ -666,6 +666,14 @@ def __call__(self, key=None, default=None):
 
         return self.get(key, default)
 
+    def __setitem__(self, k, v):
+        # Validate "publish" data member to always be boolean
+        if k == "publish":
+            msg = "\"publish\" data member has to be boolean."
+            assert isinstance(v, bool), msg
```
mottosso: Would a `TypeError` be more intuitive here?

tokejepsen: Good point! Done.

mottosso:
Sweet. I think this is quite slick. Let's give this a try, and potentially we could add validation for more types, and potentially add a plug-in-type system for end-users to validate their own types too.

mottosso:
There is one thing however.. this is a backwards breaking change, so I think we must put it behind an environment variable pre-2.0. Do you remember what the last breaking change was, that we put behind a variable? Probably safe to use them same. Then we can start gathering changes to make for a 2.0.

tokejepsen:
> this is a backwards breaking change

I was wondering whether this would qualify as breaking?
This means you have to have ```publish``` data member as boolean. But I would argue that users already have it as boolean, since that is what pyblish-qml and pyblish-lite needs.

> Do you remember what the last breaking change was, that we put behind a variable?

That would be ```PYBLISH_ALLOW_DUPLICATE_PLUGINS``` in https://github.com/pyblish/pyblish-base/releases/tag/1.6.1

mottosso:
> I was wondering whether this would qualify as breaking?

I can see what you mean, but if you put yourself in the shoes of someone updating Pyblish - not specifically for this feature, but just a general "I want the latest" - then he may discover that now things no longer work. He'll get a TypeError, even though he hasn't changed anything.

That's what makes it a backwards incompatible change.

Now, arguably, he shouldn't have used a non-Bool for this value so shouldn't be seeing this error. But that's what happens when you used to allow something and now don't. Bam!

> PYBLISH_ALLOW_DUPLICATE_PLUGINS 

That's excellent, I think it's time we step this up.

1. Let's put this change under a `PYBLISH_STRICT_DATATYPES`
2. And override both (i.e. enable both) with a `PYBLISH_EARLY_ADOPTER`

Such that enabling either variable means enabling either feature independently (for fine control) and enabling early-adopter gets you what 2.0 will eventually look like.

That way, we can keep making incompatible changes and also remove deprecated stuff, so long as it only breaks things explicitly, by setting variables such as these.

**Example**

```python
# Fine control
ALLOW_DUPLICATE_PLUGINS = bool(os.getenv("PYBLISH_ALLOW_DUPLICATE_PLUGINS"))
STRICT_DATATYPES = bool(os.getenv("PYBLISH_STRICT_DATATYPES"))

# Global override for 2.0 stuff
EARLY_ADOPTER = bool(os.getenv("PYBLISH_EARLY_ADOPTER"))
ALLOW_DUPLICATE_PLUGINS = EARLY_ADOPTER or ALLOW_DUPLICATE_PLUGINS
STRICT_DATATYPES = EARLY_ADOPTER or STRICT_DATATYPES 
```

I know this falls a little outside of this issue, we could either merge this as-is, not release it, and make another issue for these features. Or do them now. Up to you.

tokejepsen:
Think that sounds good :) Done.

pyblish/plugin.py
```diff
@@ -666,6 +671,15 @@ def __call__(self, key=None, default=None):
 
         return self.get(key, default)
 
+    def __setitem__(self, k, v):
+        # Backwards incompatible data validation.
+        if STRICT_DATATYPES:
```
mottosso, line 676: For perfect compatibility, you could put the if-statement *outside* of the method declaration. That way, the override would only happen for early adopters, avoiding potential edge-cases where overriding this causes trouble.

tests/test_plugin.py
```diff
@@ -931,3 +931,27 @@ def test_duplicate_plugin_names():
     # NOTE: This assumes the test succeeds. If it fails, then
     # subsequent tests can fail because of it.
     pyblish.plugin.ALLOW_DUPLICATES = False
+
+
+@with_setup(lib.setup_empty, lib.teardown)
+def test_validate_publish_data_member_type():
+    """Validate publish data member type works."""
+
+    pyblish.plugin.STRICT_DATATYPES = True
```
mottosso, line 940: ..Although, that would limit your ability to do this.

Let's leave it. Looks good!

mottosso:
Genius! Looking good, merge at will!

PR #326, opened 2017-10-27
Publish iterator

tokejepsen:
This PR is an effort the get an iterator for publishing for iterable feedback.

This relates to https://github.com/pyblish/pyblish-nuke/pull/36.

### DIFF
--- a/pyblish/util.py
+++ b/pyblish/util.py
@@ -97,7 +97,7 @@ def publish_iter(context=None, plugins=None, targets=None):
     for Plugin, instance in logic.Iterator(collectors, context):
         result = plugin.process(Plugin, context, instance)
         tasks_processed_count += 1
-        result["percentage"] = float(tasks_processed_count) / len(plugins)
+        result["progress"] = float(tasks_processed_count) / len(plugins)
         yield result
 
     # Exclude collectors from further processing
@@ -122,7 +122,7 @@ def publish_iter(context=None, plugins=None, targets=None):
         try:
             result = plugin.process(Plugin, context, instance)
             tasks_processed_count += 1
-            result["percentage"] = (
+            result["progress"] = (
                 float(tasks_processed_count) / (len(tasks) + len(collectors))
             )
         except StopIteration:  # End of items
--- a/pyblish/util.py
+++ b/pyblish/util.py
@@ -36,9 +36,6 @@ def publish(context=None, plugins=None, targets=None):
 
     """
 
-    # Need to generate a Context object if None is passed. This is because,
-    # if None is passed the publish iterator won't process the context and not
-    # return anything.
     context = api.Context() if context is None else context
 
     # Since the Context object is a singleton, we can just run though the
--- a/pyblish/util.py
+++ b/pyblish/util.py
@@ -54,7 +54,7 @@ def publish(context=None, plugins=None, targets=None):
 
     # First pass, collection
     for Plugin, instance in logic.Iterator(collectors, context):
-        plugin.process(Plugin, context, instance)
+        yield plugin.process(Plugin, context, instance)
 
     # Exclude collectors from further processing
     plugins = list(p for p in plugins if p not in collectors)
@@ -76,7 +76,7 @@ def publish(context=None, plugins=None, targets=None):
     for Plugin, instance in logic.Iterator(plugins, context, state):
         try:
             result = plugin.process(Plugin, context, instance)
-
+            yield result
         except StopIteration:  # End of items
             raise
 
@@ -104,7 +104,7 @@ def publish(context=None, plugins=None, targets=None):
     for target in targets:
         api.deregister_target(target)
 
-    return context
+    yield context
 
 
 def collect(context=None, plugins=None, targets=["default"]):
--- a/pyblish/util.py
+++ b/pyblish/util.py
@@ -26,10 +26,57 @@ def publish(context=None, plugins=None, targets=None):
             defaults to results of discover()
         targets (list, optional): Targets to include for publish session.
 
+    Returns:
+        Context: The context processed by the plugins.
+
     Usage:
         >> context = plugin.Context()
         >> publish(context)     # Pass..
         >> context = publish()  # ..or receive a new
+
+    """
+
+    # Need to generate a Context object if None is passed. This is because,
+    # if None is passed the publish iterator won't process the context and not
+    # return anything.
+    context = api.Context() if context is None else context
+
+    # Since the Context object is a singleton, we can just run though the
+    # publish iterator without assigning any variables.
+    for _ in publish_iter(context, plugins, targets):
+        pass
+
+    return context
+
+
+def publish_iter(context=None, plugins=None, targets=None):
+    """Publish iterator
+
+    This function will process all available plugins of the
+    currently running host, publishing anything picked up
+    during collection.
+
+    Arguments:
+        context (Context, optional): Context, defaults to
+            creating a new context
+        plugins (list, optional): Plug-ins to include,
+            defaults to results of discover()
+        targets (list, optional): Targets to include for publish session.
+
+    Yields:
+        tuple of dict and Context: A tuple is returned with a dictionary and
+            the Context object. The dictionary contains all the result
+            information of a plugin process, and the Context is the Context
+            after the plugin has been processed.
+
+    Usage:
+        >> context = plugin.Context()
+        >> for result, context in util.publish_iter(context):
+               print result
+               print context
+        >> for result, context in util.publish_iter():
+               print result
+               print context
 
     """
 
@@ -54,7 +101,7 @@ def publish(context=None, plugins=None, targets=None):
 
     # First pass, collection
     for Plugin, instance in logic.Iterator(collectors, context):
-        yield plugin.process(Plugin, context, instance)
+        yield (plugin.process(Plugin, context, instance), context)
 
     # Exclude collectors from further processing
     plugins = list(p for p in plugins if p not in collectors)
@@ -76,7 +123,7 @@ def publish(context=None, plugins=None, targets=None):
     for Plugin, instance in logic.Iterator(plugins, context, state):
         try:
             result = plugin.process(Plugin, context, instance)
-            yield result
+            yield (result, context)
         except StopIteration:  # End of items
             raise
 
@@ -103,8 +150,6 @@ def publish(context=None, plugins=None, targets=None):
     # Deregister targets
     for target in targets:
         api.deregister_target(target)
-
-    yield context
 
 
 def collect(context=None, plugins=None, targets=["default"]):
--- a/pyblish/util.py
+++ b/pyblish/util.py
@@ -125,8 +125,6 @@ def publish_iter(context=None, plugins=None, targets=None):
             result["percentage"] = (
                 float(tasks_processed_count) / (len(tasks) + len(collectors))
             )
-            yield result
-
         except StopIteration:  # End of items
             raise
 
@@ -147,6 +145,8 @@ def publish_iter(context=None, plugins=None, targets=None):
         error = result["error"]
         if error is not None:
             print(error)
+
+        yield result
 
     api.emit("published", context=context)
 
--- a/pyblish/util.py
+++ b/pyblish/util.py
@@ -80,6 +80,8 @@ def publish_iter(context=None, plugins=None, targets=None):
 
     """
 
+    percentage = 0.0
+
     # Include "default" target when no targets are requested.
     if targets is None:
         targets = ["default"]
@@ -100,8 +102,11 @@ def publish_iter(context=None, plugins=None, targets=None):
     )
 
     # First pass, collection
+    plugin_processed_count = 0
     for Plugin, instance in logic.Iterator(collectors, context):
-        yield (plugin.process(Plugin, context, instance), context)
+        plugin_processed_count += 1
+        percentage = float(plugin_processed_count) / len(plugins)
+        yield (percentage, plugin.process(Plugin, context, instance), context)
 
     # Exclude collectors from further processing
     plugins = list(p for p in plugins if p not in collectors)
@@ -120,16 +125,42 @@ def publish_iter(context=None, plugins=None, targets=None):
     }
 
     # Second pass, the remainder
+    plugin_processing = None
+    instances_processed_count = 0
+    total_plugins = len(plugins) + len(collectors)
     for Plugin, instance in logic.Iterator(plugins, context, state):
         try:
             result = plugin.process(Plugin, context, instance)
-            yield (result, context)
+
+            # Calculate percentage
+            instances_count = len(logic.instances_by_plugin(context, Plugin))
+
+            instances_processed_fraction = 0
+            if instances_count != 0:
+                instances_processed_fraction = (
+                    float(instances_processed_count) / instances_count
+                )
+
+            plugin_instance_processed = (
+                instances_processed_fraction + plugin_processed_count
+            )
+            percentage = plugin_instance_processed / total_plugins
+
+            # Yield results
+            yield (percentage, result, context)
+
+            # Increment for next iteration
+            instances_processed_count += 1
+            if plugin_processing != Plugin:
+                plugin_processed_count += 1
+                instances_processed_count = 0
+                plugin_processing = Plugin
         except StopIteration:  # End of items
             raise
 
-        except:  # This is unexpected, most likely a bug
+        except Exception as e:  # This is unexpected, most likely a bug
             log.error("An exception occurred.\n")
-            raise
+            raise e
 
         else:
             # Make note of the order at which the
--- a/pyblish/util.py
+++ b/pyblish/util.py
@@ -38,8 +38,6 @@ def publish(context=None, plugins=None, targets=None):
 
     context = api.Context() if context is None else context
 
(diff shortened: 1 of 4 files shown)


tokejepsen:
This is a simple conversion, which fails a lot of tests, but its a beginning of a discussion on how to implement this.

Initially I was thinking taking the code of the ```util.publish```  into a new method ```util.publish_iterator```. ```util.publish``` would just be a simple iteration but returning the last result (the context).

Then other methods have the option to use the ```util.publish_iterator``` for interactive feedback of the publishing process.

mottosso:
> Initially I was thinking taking the code of the util.publish into a new method util.publish_iterator. util.publish would just be a simple iteration but returning the last result (the context).

I think that sounds like a good idea. Let's call it `publish_iter` though, to align with the keyword `iter`

tokejepsen:
I opted in the end for the iterator to return a tuple of the results and the context. It felt weird for a generator to return a different type of result depending on what iteration it had reached.

I still need to write a test for the publish iterator, but wanted your opinion on this first.

tokejepsen:
What do you think of the code now?

pyblish/util.py
```diff
@@ -26,13 +26,62 @@ def publish(context=None, plugins=None, targets=None):
...
+        Context: The context processed by the plugins.
+
     Usage:
         >> context = plugin.Context()
         >> publish(context)     # Pass..
         >> context = publish()  # ..or receive a new
 
     """
 
+    # Need to generate a Context object if None is passed. This is because,
```
mottosso: I don't understand this? Can you give me an example?

mottosso:
I just realised, to get a hold of the context from the iterator, at first I was thinking it would need to be yielded every time. But then it occurred to me that, the context is mutable. If you wanted a hold of it, you could make it yourself and *pass it to the iterator*.

I think this is great, because it means less to yield. But perhaps most interestingly, it means we could potentially yield *only* the result dictionary. The only obstacle to that is the percentage, which I think we could safely include for now. At least as a test. And then consider the percentage as *optional*. As in, it doesn't need to be included in results from everywhere, like the GUIs. But it can.

We could end up with this, which I think is quite elegant.

```python
for result in util.publish_iter():
  ...

# Or..
context = api.Context()
for result in util.publish_iter(context):
  ...
```

pyblish/util.py
```diff
@@ -53,8 +102,11 @@ def publish(context=None, plugins=None, targets=None):
     )
 
     # First pass, collection
+    plugin_processed_count = 0
     for Plugin, instance in logic.Iterator(collectors, context):
-        plugin.process(Plugin, context, instance)
+        plugin_processed_count += 1
+        percentage = float(plugin_processed_count) / len(plugins)
+        yield (percentage, plugin.process(Plugin, context, instance), context)
```
mottosso: This is risky, as `plugin.process` could throw an exception?

mottosso: Ah no, sorry. I misread `plugin` for `Plugin`.

pyblish/util.py
```diff
@@ -73,16 +125,42 @@ def publish(context=None, plugins=None, targets=None):
...
+            instances_processed_count += 1
+            if plugin_processing != Plugin:
+                plugin_processed_count += 1
+                instances_processed_count = 0
+                plugin_processing = Plugin
         except StopIteration:  # End of items
             raise
 
-        except:  # This is unexpected, most likely a bug
+        except Exception as e:  # This is unexpected, most likely a bug
```
mottosso: Can remove the `as e` and just use `throw`, unless you're going to make use of the `e`. `throw` will default to re-throwing the last thrown exception.

pyblish/util.py
```diff
@@ -73,16 +125,42 @@ def publish(context=None, plugins=None, targets=None):
...
     # Second pass, the remainder
+    plugin_processing = None
+    instances_processed_count = 0
+    total_plugins = len(plugins) + len(collectors)
     for Plugin, instance in logic.Iterator(plugins, context, state):
         try:
             result = plugin.process(Plugin, context, instance)
 
+            # Calculate percentage
+            instances_count = len(logic.instances_by_plugin(context, Plugin))
```
mottosso: Do we need to be this exact...? Maybe.. You could also make `logic.Iterator` into a list, which would give you a fixed series of tasks to perform, and get the progress from `task_number / len(tasks)`

pyblish/util.py
```diff
@@ -73,16 +125,42 @@ def publish(context=None, plugins=None, targets=None):
...
+
+            instances_processed_fraction = 0
+            if instances_count != 0:
+                instances_processed_fraction = (
+                    float(instances_processed_count) / instances_count
```
mottosso: Great use of parentheses to combat line-length issues!

pyblish/util.py
```diff
@@ -73,16 +125,42 @@ def publish(context=None, plugins=None, targets=None):
...
+                instances_processed_fraction + plugin_processed_count
+            )
+            percentage = plugin_instance_processed / total_plugins
+
+            # Yield results
+            yield (percentage, result, context)
+
+            # Increment for next iteration
+            instances_processed_count += 1
+            if plugin_processing != Plugin:
```
mottosso: Yeaah, this is a little inelegant.

mottosso:
The beauty of an iterator in this sense is that you now control when the next task is due. This means you can include a `cancel` button and other interactive gestures in your GUI, like this.

```python
state["running"] = True

def on_clicked():
  state["running"] = False

button.clicked.connect(on_clicked)

for result in util.publish_iter():
  if not state["running"]:
    print("Cancelled")
    return

  print(result)
```

Or, even more interestingly, you can free yourself from having to use a loop. The problem with loops is that you can't stop them to do other things, like render the GUI. Running this, your button would do what it's supposed to, but it would only ever get a chance to update inbetween printing the result and the next plug-in starting to run. It'll appear very laggy to the user.

What you can do instead, is not use a loop.

```python
def on_next():
  if not state["running"]:
    print("Cancelled")
    return

  # This'll block the GUI, and run the next task
  result = next(iterator)

  # Run the next task, but take your time.
  # This is where the GUI gets a chance to be responsive
  # You can spend this time painting the GUI, for example, updating
  # The background color to signal success/failure, or populate
  # a list of things, maybe a textual or graphical summary of the result
  # dictionary from above. No end to the possibilities!
  QtCore.QTimer.singleShot(500, on_next)  

# Initialise publishing
context = api.Context()
iterator = util.publish_iter(context)

# Kick-start the process
on_next(iterator)
```

Notice there is no loop! Instead, each task is "chained" together, the last one calling the next. The beauty is that you can insert a delay to actually calling the next task! It's like a car engine. One ignition causing the next. This is impossible with a loop. It's the equivalent of calling `os.sleep()` in the middle of a loop, except you aren't actually blocking while sleeping. This is how you can get a GUI responsive, while processing things at the same time.

I took a moment to type this up, because the other day you asked about how to make a GUI responsive while busy with stuff. This is it! :)

pyblish/util.py
tokejepsen: If we pass in ```None``` as the context, ei. don't pass in any context to the method, the ```publish_iter``` method won't process the context object, and when we return the context object (https://github.com/pyblish/pyblish-base/pull/326/files/1306986e78f49655c131eb1787dd76d31bf4f716#diff-45fca2405e95038f14f9537f62615994R49), then it'll be ```None``` from the default value.
More specifically the two yields in ```publish_iter``` don't get executed, so the context object does not get modified. This could be resolved by always yielding the context at the end of ```publish_iter```, but then we get into the issue of yielding different results per iteration, which in my opinion makes it tricky to work with.

mottosso: Aah, ok, now I understand. Yes that makes sense, but the description could use some work. In fact, I don't think you need it, it's pretty clear you need it in order to return it.

pyblish/util.py
tokejepsen: You are right, just tested it and it works better plus simplifies the code a lot. Pushed the code for a review.

pyblish/util.py
```diff
@@ -73,16 +126,22 @@ def publish(context=None, plugins=None, targets=None):
     }
 
     # Second pass, the remainder
+    tasks = list(logic.Iterator(plugins, context, state))
     for Plugin, instance in logic.Iterator(plugins, context, state):
```
mottosso: You can use the tasks you compute here, `for Plugin, instance in tasks:`

pyblish/util.py
tokejepsen: Fair enough, I can get rid of the comment :)

pyblish/util.py
tokejepsen: Done

pyblish/util.py
tokejepsen: I actually tried that, but some tests fail:

```
FAIL: Decrementing order works fine
  File "C:\Users\tje\AppData\Local\Continuum\Miniconda2\envs\tgbvfx-environment\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\Users\tje\Documents\conda-git-deployment\repositories\tgbvfx-environment\pyblish-base\tests\pre13\test_logic.py", line 359, in test_decrementing_order
    assert_equals(count["#"], 111.1)
AssertionError: 1111.1 != 111.1
-------------------- >> begin captured stdout << ---------------------
I will run first, and stop things
I will stop things
I will not run

--------------------- >> end captured stdout << ----------------------

FAIL: The test halts an invalid publish
  File "C:\Users\tje\AppData\Local\Continuum\Miniconda2\envs\tgbvfx-environment\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\Users\tje\Documents\conda-git-deployment\repositories\tgbvfx-environment\pyblish-base\tests\pre13\test_logic.py", line 382, in test_test
    assert count["#"] == 11
AssertionError:
-------------------- >> begin captured stdout << ---------------------
I will run first, and stop things
I will not run

--------------------- >> end captured stdout << ----------------------

FAIL: ContextPlugin works as advertised
  File "C:\Users\tje\AppData\Local\Continuum\Miniconda2\envs\tgbvfx-environment\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\Users\tje\Documents\conda-git-deployment\repositories\tgbvfx-environment\pyblish-base\tests\test_plugin.py", line 563, in test_explicit_plugin
    assert count["#"] == 111, count
AssertionError: {'#': 1111}
-------------------- >> begin captured stdout << ---------------------
fail

--------------------- >> end captured stdout << ----------------------
-------------------- >> begin captured logging << --------------------
pyblish.Collector: INFO: Collecting from ContextPlugin
--------------------- >> end captured logging << ---------------------

```

I don't quite understand it, but I think its because the iterator raises an exception https://github.com/pyblish/pyblish-base/blob/477f540981433867aa219cc46cd0fa5890ecf17c/pyblish/logic.py#L372

tokejepsen:
> I think this is great, because it means less to yield. But perhaps most interestingly, it means we could potentially yield only the result dictionary. The only obstacle to that is the percentage, which I think we could safely include for now. At least as a test. And then consider the percentage as optional. As in, it doesn't need to be included in results from everywhere, like the GUIs. But it can.

Sure, I'll take a look at this.

> I took a moment to type this up, because the other day you asked about how to make a GUI responsive while busy with stuff. This is it! :)

Thanks for taking the time on that post :) Very interesting!

pyblish/util.py
mottosso: Ah, yes you are right.

This is another thing about iterators, while running, you can change the conditions within which it considers the next item to yield. In this case, your percentage won't be accurate. It may be "close enough", you'll have to experiment.

tokejepsen: > It may be "close enough", you'll have to experiment.

I might not be understanding the processing flow correctly, but in order to get an accurate percentage I would have to predict the flow which means running the ```process``` methods twice?
I don't really mind that the percentage isn't fully accurate. You could also argue that if a validator fails, the progress bar will stop. Dunno if that is the same scenario.

tokejepsen:
Are we getting to a mergable state?

pyblish/util.py
```diff
@@ -26,13 +26,59 @@ def publish(context=None, plugins=None, targets=None):
...
     Usage:
         >> context = plugin.Context()
         >> publish(context)     # Pass..
         >> context = publish()  # ..or receive a new
 
     """
 
+    context = api.Context() if context is None else context
+
+    # Since the Context object is a singleton, we can just run though the
```
mottosso: Oh, it's not a singleton. Every context is unique. There's no need for this comment either IMO.

pyblish/util.py
```diff
@@ -26,13 +26,59 @@ def publish(context=None, plugins=None, targets=None):
...
+    Yields:
+        tuple of dict and Context: A tuple is returned with a dictionary and
+            the Context object. The dictionary contains all the result
+            information of a plugin process, and the Context is the Context
+            after the plugin has been processed.
+
+    Usage:
+        >> context = plugin.Context()
+        >> for result, context in util.publish_iter(context):
+               print result
```
mottosso: Could you please make the print-statements Python 3 friendly? `print(result)`

pyblish/util.py
tokejepsen: Done :)

pyblish/util.py
tokejepsen: And done :)

pyblish/util.py
```diff
@@ -73,9 +117,15 @@ def publish(context=None, plugins=None, targets=None):
...
     # Second pass, the remainder
+    tasks = list(logic.Iterator(plugins, context, state))
     for Plugin, instance in logic.Iterator(plugins, context, state):
         try:
             result = plugin.process(Plugin, context, instance)
+            tasks_processed_count += 1
+            result["percentage"] = (
+                float(tasks_processed_count) / (len(tasks) + len(collectors))
+            )
+            yield result
```
mottosso: Better yield at the very end of the loop. If you look below, you can see how unexpected exceptions are handled. At the moment, it would yield what is supposed to be a proper result before reaching that check, and yield an exception.

tokejepsen: Good point.. Done :)

mottosso:
Looks good to me! If you increment the (minor) version on this, you're welcome to make a release. If possible, it would be fantastic to include a little example in the release of how and when to use it.

tokejepsen:
> Looks good to me! If you increment the (minor) version on this, you're welcome to make a release. If possible, it would be fantastic to include a little example in the release of how and when to use it.

Should we cement the behaviour of ```publish_iter``` with tests, or keep it flexible while we figure out how its being used?

mottosso:
Oooh nice one. I would like to say I was testing you, not enforcing tests, but I hadn't even considered it. Yes, you are right, this definitely needs tests.

But, I also think you're right about us needing to figure out how we would like to use it. If you could use this to build your GUI, then that would be a good start. Then we can have a look at adding tests to keep the behaviour consistent in the future.

tokejepsen:
Alright, let's continue the discussion at the pyblish-nuke side and once we are happy with that PR, I'll make some tests for this to make sure the behaviour is consistent in the future.

tokejepsen:
Added the most basic tests I could think of, that needs to be in place for pyblish-nuke to function.

I'm hesitant to make more detailed tests about how the specifically the progress calculation works as we might change or update this as we figure out how its used.

What do you think?

mottosso:
We need tests for..

1. What happens when there are no plug-ins?
2. What happens when a bad plug-in is registered?
3. What happens when one iterator is called from two threads?

What other edge cases can you think of?

tokejepsen:
> What other edge cases can you think of?

Sorry, drawing a blank for this.

I'll have a look those requested tests.

mottosso:
The threads one applies to the vanilla `publish()` too, but I think in case of the iterator, it's more likely to be used in a GUI and one might think to use it from a thread to keep the GUI running. In Maya, this would crash. So maybe we could add to `publish()` a check that it's being called from the main thread.

tokejepsen:
> What happens when there are no plug-ins?

What do you expect to test with this?
When there are no plugin-ins, the iterator is just going to process nothing and any optional context passed in will be left empty.

> What happens when a bad plug-in is registered?

What do you define as a bad plug-in? Syntax issues?

tokejepsen:
> What happens when one iterator is called from two threads?

Very unsure about how to tackle this. Sounds like you want it to be tested in the hosts?

mottosso:
No problem, I'll have a closer look now.

mottosso:
Putting together something with the new features, and pushed a few things I found missing. Let me know what you think.

tokejepsen:
All looks good to me.

Got a goal in mind with these changes?

mottosso:
> Got a goal in mind with these changes?

Yes, we need an example to fully validate the need for it, so I'll put together a small GUI (a little more than a splash image, but not much). As I was writing it, I found I needed `collect()` and friends, but they weren't there. I think others will come to expect it to be there too.

tokejepsen:
Cool, let me know when you have something to look at.

mottosso:
This is a good feature and is well implemented. It's delayed because it opens so many new possibilities it'd be a shame not to document at least some of them and work up an example, but seeing as I haven't gotten around to doing that yet I figure it'd do more good out in the wild than sitting in here.

So, merging!

PR #327, opened 2017-11-21
Check whether any registered guis override default

iwootten:
**Motivation**

Allow the cli to use registered guis set by PYBLISH_GUI. Currently when the environment var has been set, pyblish still needs to be called supplying it as an argument.

**Changes**

This change looks for any registered guis before falling back to the default package argument.


tokejepsen:
Working here :)

PR #328, opened 2017-11-30
plugin path normalize

davidlatwe:
Two simple improvement as follow :
1. use `os.path.join` instead of hard-coded slash to compose default plugin path.
2. use `os.path.normpath` to normalize plugin path before register/deregister it.

### DIFF
--- a/pyblish/api.py
+++ b/pyblish/api.py
@@ -152,7 +152,7 @@ def __init__():
             register_host(host)
 
     # Register default path
-    register_plugin_path("%s/plugins" % __main_package_path())
+    register_plugin_path(os.path.join(__main_package_path(), "plugins"))
 
     # Register default test
     register_test(__default_test)
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -1008,10 +1008,12 @@ def register_plugin_path(path):
     Example:
         >>> import os
         >>> # Linux
-        >>> register_plugin_path("server/pluginsA") == os.path.join("server", "pluginsA")
+        >>> plugins_A = "server/pluginsA"
+        >>> register_plugin_path(plugins_A) == os.path.normpath(plugins_A)
         True
         >>> # Windows
-        >>> register_plugin_path(r"server\pluginsB") == os.path.join("server", "pluginsB")
+        >>> plugins_B = r"server\pluginsB"
+        >>> register_plugin_path(plugins_B) == os.path.normpath(plugins_B)
         True
         >>> deregister_plugin_path("server/pluginsA")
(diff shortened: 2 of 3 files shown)


davidlatwe:
Ummm, not familiar with this.
Don't know why travis-ci failing the doctest but appveyor succeeded.

Was trying to make the tests to except the normalized path as result.

mottosso:
Travis is Linux, AppVeyor is Windows.

I've updated the doctest, this might work.

mottosso:
Ah! Doi. My bad. Having another look. :)

davidlatwe:
This gonna work !

mottosso:
Looks like we've found a bug. Deregister will need normpath as well.

davidlatwe:
But I did add it there, too :(

davidlatwe:
Turns out that, in Windows
```python
>>> os.path.normpath("server/pluginsB")
'server\\pluginsB'
>>> os.path.normpath(r"server\pluginsB")
'server\\pluginsB'
```

but in Linux
```python
>>> os.path.normpath("server/pluginsB")
'server/pluginsB'
>>> os.path.normpath(r"server\pluginsB")
'server\\pluginsB'
```
phew...

mottosso:
Ah, but are you sure `normpath` is what you are looking for then? I think what we're looking for is something to replace any backslash with the right one for the platform. Or at least a similar backslash such that you can register/deregister without thinking about it. This doesn't solve anything?

davidlatwe:
>  I think what we're looking for is something to replace any backslash with the right one for the platform.

The `normpath` did make Windows platform easier passing path, no matter backslash or not, however Linux could use backslash as part of folder name...

> This doesn't solve anything?

Me personally think that `normpath` could be like a simple guard for `_registered_paths`. But assume most of the case we just copy the path which registered previously, and paste to deregister, so maybe this is unnecessary, only the first commit in this PR is needed. :(

davidlatwe:
How about not to do `normpath` in this PR, but making `deregister_plugin_path()` to print out what path it tried to remove when failed ? Not just return `x not in list` error.
~current registered paths if failed to remove one ?~

Easier to debug if in an automation process.

davidlatwe:
It's off this PR's original purpose now, it was at first try to normalize folder path's separator for plugin path's registration more straight forward, but turns out it doesn't have a simple straight answer when dealing cross platform, at least I can't think of any reliable way.

Secondly, I believe that this wasn't Pyblish's responsibility need to take care of, just making more convenient, so if it doesn't have a simple answer, then this PR has failed.

I will close this PR, and make it cleaner. Sorry about that. :sneezing_face:

PR #329, opened 2017-12-01
use os.path.join to compose default plugin path

davidlatwe:
Change to `os.path.join` for cross-platform.

### DIFF
--- a/pyblish/api.py
+++ b/pyblish/api.py
@@ -152,7 +152,7 @@ def __init__():
             register_host(host)
 
     # Register default path
-    register_plugin_path("%s/plugins" % __main_package_path())
+    register_plugin_path(os.path.join(__main_package_path(), "plugins"))
 
     # Register default test
     register_test(__default_test)
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -1007,19 +1007,20 @@ def register_plugin_path(path):
 
     Example:
         >>> import os
-        >>> my_plugins = "/server/plugins"
-        >>> register_plugin_path(my_plugins)
-        '/server/plugins'
+        >>> my_plugins = os.path.join("server", "plugins")
+        >>> register_plugin_path(my_plugins) == os.path.normpath(my_plugins)
+        True
 
     Returns:
         Actual path added, including any post-processing
 
     """
 
-    if path in _registered_paths:
+    normpath = os.path.normpath(path)
+    if normpath in _registered_paths:
         return log.warning("Path already registered: {0}".format(path))
 
-    _registered_paths.append(path)
+    _registered_paths.append(normpath)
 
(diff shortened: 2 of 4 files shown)


davidlatwe:
Sham on me, wasn't think this right.

If changes to use `os.path.join` for composing default plugin path, then will need to have `os.path.normpath` to do path normalize, for someone already using backslash for default path's register/deregister (which is working but will not after this PR).

> [but are you sure normpath is what you are looking for then?](https://github.com/pyblish/pyblish-base/pull/328#issuecomment-348193381)

Now I think it's "Yes" for sure, both `/` and `\` could work as file path separator on Windows, so **it did need** `normpath` to make sure that these two file path *strings*
```python
path_1 = "server/plugins"
path_2 = r"server\plugins"
```
are the same *workable* path.

But on Linux, only `/` could use as file path separator, `\` could be part of the file or folder name, so those two file path *strings* above **won't** be recognized as same *workable* path. First one is a folder path for sure, but the second one could be a file name or else, so it's already a false input when on Linux.

> Or at least a similar backslash such that you can register/deregister without thinking about it. 

I think this did solve for Windows, but wasn't needed for Linux.

<details>
<summary>Could verify with this</summary>

```python
import os

input_1 = "server/plugin"
input_2 = r"server\plugin"

path_1 = os.path.join(os.getcwd(), os.path.normpath(input_1))
path_2 = os.path.join(os.getcwd(), os.path.normpath(input_2))
try:
    os.makedirs(path_1)
    print("path_1 created.")
except OSError:
    print("path_1 already exists.")
try:
    os.makedirs(path_2)
    print("path_2 created.")
except OSError:
    print("path_2 already exists.")

# Windows
# path_1 created.
# path_2 already exists.

# Linux
# path_1 created.
# path_2 created.

```

</details>
<br>

Then, in the Doctest, maybe there won't need two platform type of tests ? Since we won't need to test this `r"server\plugins"` on a Linux machine, because it's not a file path for Linux at all.

So I changed as follow
```python
"""
>>> import os
>>> my_plugins = os.path.join("server", "plugins")
>>> register_plugin_path(my_plugins) == os.path.normpath(my_plugins)
True
"""
```

Does this make sense ?

mottosso:
> Then, in the Doctest, maybe there won't need two platform type of tests ? Since we won't need to test this r"server\plugins" on a Linux machine, because it's not a file path for Linux at all.

Yes, I think you're right. When you register these, you should register them using the slash of the active platform, e.g. via `os.path.join()`. In which case this will work just fine.

I think this is good, it's definitely an improvement. Happy to merge this, if you could bump this to 1.5.6 in version.py then I'd be happy to merge this!

davidlatwe:
Thanks !
Right away :)

PR #332, opened 2018-01-23
Duplicate plugin names should be supported.

tokejepsen:
This is just the test to show that duplicate plugin names are not supported, so the tests should fail.

I'm pretty sure the issue is [here](https://github.com/pyblish/pyblish-base/blob/master/pyblish/plugin.py#L1287), because we compare the plugin names rather than their class string:
```
<class 'c:\users\tje\appdata\local\temp\tmponhxwt\pluginA.py.collectorA'>
```

What I could not find was where we set this string?

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -1239,6 +1239,7 @@ def discover(type=None, regex=None, paths=None):
                       "has been deprecated and does nothing")
 
     plugins = dict()
+    plugin_names = []
 
     # Include plug-ins from registered paths
     for path in paths or plugin_paths():
@@ -1277,6 +1278,19 @@ def discover(type=None, regex=None, paths=None):
                 continue
 
             for plugin in plugins_from_module(module):
+                # Check for duplicate plugin names. This is to preserve
+                # backwards compatility.
+                allow_duplicates = eval(
+                    os.environ.get(
+                        "PYBLISH_ALLOW_DUPLICATE_PLUGIN_NAMES", "False"
+                    )
+                )
+                if not allow_duplicates:
+                    if plugin.__name__ in plugin_names:
+                        log.debug("Duplicate plug-in found: %s", plugin)
+                        continue
+                    plugin_names.append(plugin.__name__)
+
                 plugin.__module__ = module.__file__
                 key = "{0}.{1}".format(plugin.__module__, plugin.__name__)
                 plugins[key] = plugin
@@ -1284,6 +1298,19 @@ def discover(type=None, regex=None, paths=None):
     # Include plug-ins from registration.
     # Directly registered plug-ins take precedence.
     for plugin in registered_plugins():
+        # Check for duplicate plugin names. This is to preserve
+        # backwards compatility.
+        allow_duplicates = eval(
+            os.environ.get(
+                "PYBLISH_ALLOW_DUPLICATE_PLUGIN_NAMES", "False"
+            )
+        )
+        if not allow_duplicates:
+            if plugin.__name__ in plugin_names:
+                log.debug("Duplicate plug-in found: %s", plugin)
+                continue
+            plugin_names.append(plugin.__name__)
+
         plugins[plugin.__name__] = plugin
 
     plugins = list(plugins.values())
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -1285,11 +1285,10 @@ def discover(type=None, regex=None, paths=None):
                         "PYBLISH_ALLOW_DUPLICATE_PLUGIN_NAMES", "False"
                     )
                 )
-                if not allow_duplicates:
-                    if plugin.__name__ in plugin_names:
-                        log.debug("Duplicate plug-in found: %s", plugin)
-                        continue
-                    plugin_names.append(plugin.__name__)
+                if not allow_duplicates and plugin.__name__ in plugin_names:
+                    log.debug("Duplicate plug-in found: %s", plugin)
+                    continue
+                plugin_names.append(plugin.__name__)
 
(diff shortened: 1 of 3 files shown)


mottosso:
Hm, should this really be supported? The thing we implemented support for last time around was duplicate *file names*.

tokejepsen:
> Hm, should this really be supported? The thing we implemented support for last time around was duplicate file names.

The use-case is when you have lots of plugins from different sources, it can be tricky to get a unique name that is descriptive and short.

tokejepsen:
The logic for supporting duplicate names are there, but ```api.discover()``` filters duplicate plugin names.

mottosso:
Ok, welcome to remove that check. If the tests hold, then I can't see why not.

tokejepsen:
There is a single failing test, which is a direct contradiction to what we are trying to achieve here: https://github.com/pyblish/pyblish-base/blob/b9535371362f480dfe00740de1af553870c8cdd4/tests/pre11/test_plugins.py#L42-L54

This a test for pre11, so the question is how to handle this?
We could remove the test, but that'll break backwards compatibility. We could also have a check for duplicate plugin names for pre11 versions only.

mottosso:
Ok, in that case we'll need to make it opt-in.

Such as an environment var.

```bash
$ set PYBLISH_ALLOW_DUPLICATE_PLUGINS=True
```

Then for 2.0 we could make it default. Not ideal, nor clean. But backwards compatibility is king.

tokejepsen:
I opted for ```PYBLISH_ALLOW_DUPLICATE_PLUGIN_NAMES``` in the end.

The code might be a little verbose, so let me know what you think.

pyblish/plugin.py
```diff
@@ -1277,19 +1278,39 @@ def discover(type=None, regex=None, paths=None):
                 continue
 
             for plugin in plugins_from_module(module):
-                if plugin.__name__ in plugins:
-                    log.debug("Duplicate plug-in found: %s", plugin)
-                    continue
+                # Check for duplicate plugin names. This is to preserve
+                # backwards compatility.
+                allow_duplicates = eval(
```
mottosso: What's `eval` for?

I'd suggest..

```python
ALLOW_DUPLICATES = bool(os.getenv("PYBLISH_ALLOW_DUPLICATE_PLUGIN_NAMES"))
```

At the header of the module and..

```python
                    if not ALLOW_DUPLICATES and plugin.__name__ in plugin_names:
                        log.debug("Duplicate plug-in found: %s", plugin)
                        continue
```

As a change to the body.

tokejepsen:
> What's eval for?

```allow_duplicates = bool(os.getenv("PYBLISH_ALLOW_DUPLICATE_PLUGIN_NAMES"))``` does not evaluate the string to a boolean. For example ```bool("False")``` is ```True```.

File: pyblish/plugin.py (hunk 1 above)
Comment (tokejepsen) [reply pyblish/plugin.py]: > What's eval for?

```allow_duplicates = bool(os.getenv("PYBLISH_ALLOW_DUPLICATE_PLUGIN_NAMES"))``` does not evaluate the string to a boolean. For example ```bool("False")``` is ```True```.

Comment (mottosso) [reply pyblish/plugin.py]: Aah, ok, no that's fine. If the variable exists, no matter its value, then it's `True`. Common practice.

**Comment by tokejepsen** [comment]
Ready to merge?

**Comment by mottosso** [comment]
No, need to get rid of the eval stuff.

**Comment by tokejepsen** [comment]
> No, need to get rid of the eval stuff.

Thought I did?

File: pyblish/plugin.py [hunk 2]
```diff
@@ -1284,19 +1285,33 @@ def discover(type=None, regex=None, paths=None):
                 continue
 
             for plugin in plugins_from_module(module):
-                if plugin.__name__ in plugins:
+                # Check for duplicate plugin names. This is to preserve
+                # backwards compatility.
+                allow_duplicates = bool(
```
mottosso: This can be evaluated on module-import, no need to do it every iteration of this loop.

tokejepsen: The issue with putting it at the module level, is that the tests can't change the behaviour per test.

tokejepsen: How about the latest?
Its once at the function level.

mottosso: Ah, I see what you mean. Well, you can always do `module.THIS_VARIABLE = True` for testing's sake. I think that should be fine.

tokejepsen: Good point. Done.

mottosso:
Ok, I'm happy. Merge when ready. 👍

RafaelVillar:
Hello, in regards of the duplicate plugins of the same name. 
Doesn't the below always default to True no matter the string?

`
ALLOW_DUPLICATES = bool(
    os.getenv("PYBLISH_ALLOW_DUPLICATE_PLUGIN_NAMES")
)
`

True = bool("blah"), bool("false"), bool("False") 

vs

`
ALLOW_DUPLICATES = ast.literal_eval(
    os.getenv("PYBLISH_ALLOW_DUPLICATE_PLUGIN_NAMES")
)
`
False = bool("False")
True = bool("True")

tokejepsen:
Hey @RafaelVillar 

You are right. Once the environment variable ```PYBLISH_ALLOW_DUPLICATE_PLUGIN_NAMES``` is available, no matter what that variable is, it'll evaluate as True.

mottosso:
That's right, any value == True. It's by design and aligns with how boolean environment variables are implemented elsewhere, like Maya.

RafaelVillar:
Thank you for pointing that out. I suppose I didnt think of it like that. Not my initial preference, but good to know.

PR #333, opened 2018-02-02
Fix restore state of duplicate plugin test.

tokejepsen:
Pretty sure we meant for the state of duplicate plugins to be restored to ```False```.

### DIFF
--- a/tests/test_plugin.py
+++ b/tests/test_plugin.py
@@ -926,9 +926,8 @@ def test_duplicate_plugin_names():
         plugins.extend(pyblish.api.discover(paths=[temp]))
 
     assert len(plugins) == 2, plugins
-    
+
     # Restore state, for subsequent tests
     # NOTE: This assumes the test succeeds. If it fails, then
     # subsequent tests can fail because of it.
-    pyblish.plugin.ALLOW_DUPLICATES = True
-
+    pyblish.plugin.ALLOW_DUPLICATES = False


mottosso:
True or False doesn't matter, it's only the existence of this variable that matters. I shouldn't have used `True` as an example, I can see how that made things confusing. It's the variable that's saying "Yes, allow duplicates". If it isn't saying it, then duplicates aren't allowed.

So in this case, it *is* confusing because now the variable exists, but its value is "False". Does that mean duplicates *are* allowed, or not? :)

To my mind, the value can only be positive, but maybe it's a better option to make it "Yes" or "1" or anything that doesn't imply to the user that he can or should set it to anything specific.

tokejepsen:
> Does that mean duplicates are allowed, or not?

It would mean that the duplicates aren't allowed.

Are you confusing the environment variable with the actual boolean variable? 
https://github.com/pyblish/pyblish-base/blob/5f2661097dfb7c4a6dffa20e5d251b4d5beeb660/pyblish/plugin.py#L48-L52

mottosso:
Ah, yes. Sorry, you're right. But in that case, how come the tests still pass when you change this value? :O

mottosso:
Ok, I see what's going on. Should have paid more attention. :) This looks good.

tokejepsen:
Would this require a release?

mottosso:
No, don't think so.​

Issue #334, opened 2018-02-15
CLI missing argument for targets


aardschok:
## Issue:
We have setup a publishing approach which uses `targets` so we only get those plugins when the `target` is set to `X`. When attempting to run `pyblish publish --targets X` we are notified that there is no such option as --targets.

After looking in the code we found that the publish function has the keyword argument `targets` but the CLI module has no integration of it.

## Motivation
Update CLI to support targets

tokejepsen:
👍

aardschok:
When going over the amount of kwargs which needed to be set I tested a more global approach thanks to @BigRoy.  

Instead following the trail of `targets` throughout `pyblish-base` and `pyblish-qml` I have introduced an environment variable called `PYBLISH_TARGETS` which is being set in our `Avalon` CLI. 

Snippet:
```python
if kwargs.targets:
    os.environ["PYBLISH_TARGETS"] = os.pathsep.join(kwargs.targets)
```

In the `api` module of `pyblish-base` I ensured the environment variable is being checked and process in the `__init__` function with the following snippet:
```python
# Register targets for current session
for t in os.environ.get("PYBLISH_TARGETS", "").split(os.pathsep):
    if t:
        register_target(t)
```

I did however implement a `click.option` for `targets` to be passed on the `publish` function to ensure that, when the flag `--nogui` is used, you are still able to pass on the `targets`. 

Any thoughts on this approach?

aardschok:
Full code can be viewed here:
https://github.com/aardschok/pyblish-base/tree/334

PR #335, opened 2018-02-16
Add --targets to CLI

aardschok:
Solves #334
* Added `--targets` through `publish`
* Added check for `PYBLISH_TARGETS` env variable  
if targets are found register them with the API
* Minor cosmetics

### DIFF
--- a/pyblish/api.py
+++ b/pyblish/api.py
@@ -151,6 +151,11 @@ def __init__():
         if host:
             register_host(host)
 
+    # Register targets for current session
+    for t in os.environ.get("PYBLISH_TARGETS", "").split(os.pathsep):
+        if t:
+            register_target(t)
+
     # Register default path
     register_plugin_path(os.path.join(__main_package_path(), "plugins"))
 
--- a/pyblish/api.py
+++ b/pyblish/api.py
@@ -148,13 +148,17 @@ def __init__():
 
     # Register hosts from environment "PYBLISHHOSTS"
     for host in os.environ.get("PYBLISH_HOSTS", "").split(os.pathsep):
-        if host:
-            register_host(host)
+        if not host:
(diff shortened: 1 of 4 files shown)


tokejepsen:
Looks good to me. Maybe we need some tests for this?

aardschok:
I'll setup a test for `targets`

tokejepsen:
Looks good to me. @mottosso ?

pyblish/api.py
```diff
@@ -151,6 +151,11 @@ def __init__():
         if host:
             register_host(host)
 
+    # Register targets for current session
+    for t in os.environ.get("PYBLISH_TARGETS", "").split(os.pathsep):
```
mottosso: This is too short of a variable name.

Can we also make it..

```python
if not target:
  continue

register_target(target)
```

To reduce indentation?

mottosso: Alternatively..

```python
register_target(target) if target else None
```

mottosso:
Ok, with that minor tweak I'm happy.

pyblish/api.py
aardschok: If I do that then I want to change the lines (line 150) above it as well to have consistency.
I used this as a template to ensure the logic would be the same:
```python
# Register hosts from environment "PYBLISHHOSTS"
for host in os.environ.get("PYBLISH_HOSTS", "").split(os.pathsep):
    if host:
        register_host(host)
```

These lines would then become:
```python
# Register hosts from environment "PYBLISHHOSTS"
for host in os.environ.get("PYBLISH_HOSTS", "").split(os.pathsep):
    if not host:
        continue
    register_host(host)
```

Agreed?

mottosso:
Thanks!

PR #342, opened 2019-06-04
Feature/pype 348 perspective in lite

iLLiCiTiT:
- exception is stored to result so traceback can be formatted
- fixed list to tuple in plguin paths


iLLiCiTiT:
Sorry, I misclicked...

Issue #343, opened 2019-06-20
Custom logic for filtering discovered plugins


PaulSchweizer:
**Disclaimer**

I only started looking into pyblish two days ago and really liking it! This however means I might not be aware of any existing solutions or concepts that might already handle this request. If they exist I'd be happy about any hint in the right direction. 

**Goal**

Provide a way to use arbitrary, custom logic to influence/filter the list of discovered plugins.    

**Motivation**

When working on mutliple projects at once, each with its own requirements of what and how to publish things, the exisiting way of guiding the flow of publishing through plugin paths, hosts, families etc is not enough and also not really applicable I think. 
In studios, this problem is usually handled through a system of configuration files that can be specified per project/shot etc. 

This feature request would provide an entry point for such a configuration workflow, thus allowing studios to influence the publishing process in whatever way they want.  

**Suggested implementation**

I tested the following one-line approach successfully but maybe there are implications that I am not aware of.

One line change to: https://github.com/pyblish/pyblish-base/blob/master/pyblish/plugin.py#L1330 to just emit a new "pluginsDiscovered" signal:.

```python

def discover(type=None, regex=None, paths=None):

    [...]

    plugins = list(plugins.values())
    sort(plugins)  # In-place

    lib.emit("pluginsDiscovered", plugins=plugins)  # New signal 

    return plugins

```

Then a user could just modify the list of plugins in place, like this:

```python

def filter_plugins(plugins):

    config = ["MyValidator", "MyExtractor"]

    for plugin in reversed(plugins):
        if plugin.__name__ not in config:
            plugins.pop(plugin.index(plugin))


api.register_callback("pluginsDiscovered", filter_plugins)

```

I hope that makes sense.

mottosso:
Hi @PaulSchweizer, welcome to Pyblish and thanks for the kind words!

That's a neat feature; I think typically what folks do is leverage PYBLISHPLUGINPATH or `api.register_plugin_path`, which works a lot like PYTHONPATH. That way, you can customise your path up-front, and in this way guide discovery.

For example, for per-project or per-shot plug-ins, you could say:

```bash
$ export PYBLISHPLUGINPATH=/projects/alita/shots/1000/pyblish_plugins
$ maya
```

And from there, Pyblish would go and look for plug-ins in this directory, which is relative a given project. You can do this either before or during a run of a DCC.

[...]

PR #344, opened 2019-06-21
Implemented #343: Custom logic for filtering discovered plugins

antirotor:
### This implements #343 

Adding ability to filter plugins and modify plugins during discovery. This is using same `register_*` logic as already existing api. 

### Adding filter function:

```python
pyblish.api.register_discovery_filter(my_filter)
```
Filter function must return tuple `(plugin, filtered)` -> `(class, bool)`
**Plugin** is plugin class processed by filter function, **filtered** is bool indicating wheter plugin should be filtered out or not.

So simple filter function can look like:
```python
def my_filter_function(plugin):
    if plugin.__name__ == 'FilteredPlugin':
        return plugin, True
   return plugin, False
```
This can also modify plugin:
```python
def my_filter_function(plugin):
    if plugin.__name__ == 'ModifiedPlugin':
        plugin.optional = False
   return plugin, False
```

### This PR adds:
```
register_discovery_filter,
deregister_discovery_filter,
deregister_all_discovery_filters,
registered_discovery_filters
```

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -1199,14 +1199,15 @@ def register_discovery_filter(callback):
     """Register a new plugin filter
 
     Arguments:
-        callback (func): Function to execute on filter during discovery
+        callback (func): Function to execute on filter during discovery,
+            takes the original of plugins to be edited in-place
 
     Raises:
         ValueError if `callback` is not callable.
 
     """
 
-    if not hasattr(callback, "__call__"):
+    if not callable(callback):
         raise ValueError("%s is not callable" % callback)
 
     _registered_plugin_filters.append(callback)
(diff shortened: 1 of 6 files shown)


mottosso:
Whop, you are fast. Commented [here](https://github.com/pyblish/pyblish-base/issues/343#issuecomment-504491701).

Let's continue implementation-specifics here.

pyblish/plugin.py
```diff
@@ -31,12 +31,14 @@
...
     _registered_paths,
     _registered_targets,
+    _registered_plugin_filters
 )
 
 from . import lib
 from .vendor import iscompatible, six
 
 log = logging.getLogger("pyblish.plugin")
+from pprint import pprint
```
mottosso: Don't forget to remove unused imports.

antirotor:
So now basic filter example looks like this:

```python
def my_plugin_filter(plugins):
        for name, plugin in list(plugins.items()):
            if plugin.__name__ == "MyFilteredPlugin":
                del plugins[name]

        pass
```
Is like this because in python 3, you cannot delete from dictionary while iterating over it.

pyblish/plugin.py
```diff
@@ -1326,6 +1390,7 @@ def discover(type=None, regex=None, paths=None):
 
         plugins[plugin.__name__] = plugin
 
+    filter_plugins(plugins)
```
mottosso: I think we should move this till the very last step; after conversion to a list, and after having been sorted. That would give the user final control over the items, including their order.

pyblish/plugin.py
```diff
@@ -1194,6 +1195,69 @@ def registered_targets():
...
+
+    """
+
+    if not _registered_plugin_filters:
+        return
+
+    filtered = False
+    for callback in _registered_plugin_filters:
+        try:
+            plugin, filtered = callback(plugins)
```
mottosso: There's no need for the callback to return anything; it can change the items in-place, like they do in your test below.

mottosso:
@antirotor Thanks for this, would you mind if I made a few tweaks to your implementation?

mottosso:
Let me know what you think about these changes, and whether they still apply to your usecase.

antirotor:
Seems ok for me, looks cleaner. I have no problem with those changes.

mottosso:
Great, then we're all set. :) Thanks @antirotor

mottosso:
Released as [1.8.0](https://github.com/pyblish/pyblish-base/releases/tag/1.8.0)

PR #347, opened 2019-08-06
Better exception and traceback handling

mkolar:
These are some simple addition to the exception handling in pyblish that we required to make new pyblish-lite work (PR https://github.com/pyblish/pyblish-lite/pull/109 ). Look at them in the same context if possible :)

It has mostly to do with storing, formatting and nicely returning tracebacks so they can be displayed to user in GUIs.

### DIFF
--- a/pyblish/cli.py
+++ b/pyblish/cli.py
@@ -355,7 +355,7 @@ def gui(ctx, package):
 
     with _cli_plugin(data=context.data) as plugin_path:
         environ["PYBLISHPLUGINPATH"] = os.pathsep.join(
-            ctx.obj["plugin_paths"] + tuple(plugin_path)
+            ctx.obj["plugin_paths"] + (plugin_path, )
         )
 
         process = subprocess.Popen(
--- a/pyblish/cli.py
+++ b/pyblish/cli.py
@@ -355,7 +355,7 @@ def gui(ctx, package):
 
     with _cli_plugin(data=context.data) as plugin_path:
         environ["PYBLISHPLUGINPATH"] = os.pathsep.join(
-            ctx.obj["plugin_paths"] + [plugin_path]
+            ctx.obj["plugin_paths"] + tuple(plugin_path)
         )
 
         process = subprocess.Popen(
--- a/pyblish/lib.py
+++ b/pyblish/lib.py
@@ -57,8 +57,18 @@ def extract_traceback(exception):
     """Inject current traceback and store in exception"""
     exc_type, exc_value, exc_traceback = sys.exc_info()
     exception.traceback = traceback.extract_tb(exc_traceback)[-1]
-    del(exc_type, exc_value, exc_traceback)
 
+    trc_lines = traceback.format_exception(exc_type, exc_value, exc_traceback)
+    fname, line_no, func, exc = exception.traceback
+    error_info = {
+        'msg': str(exception),
+        'filename': str(fname),
+        'lineno': str(line_no),
+        'func': str(func),
+        'traceback': ''.join(trc_lines)
+    }
+    del(exc_type, exc_value, exc_traceback)
+    return error_info
 
 def time():
     """Return ISO formatted string representation of current UTC time."""
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -491,6 +491,7 @@ def __explicit_process(plugin, context, instance=None, action=None):
         "instance": instance,
         "action": action,
         "error": None,
+        "error_info": {},
         "records": list(),
         "duration": None,
         "progress": 0,
@@ -522,9 +523,10 @@ def __explicit_process(plugin, context, instance=None, action=None):
         # http://stackoverflow.com/a/11417308/478949
         lib.emit("pluginFailed", plugin=plugin, context=context,
                  instance=instance, error=error)
-        lib.extract_traceback(error)
+
+        error_info = lib.extract_traceback(error)
         result["error"] = error
-        result["error_info"] = sys.exc_info()
+        result["error_info"] = error_info
 
     __end = time.time()
 
@@ -558,6 +560,7 @@ def __implicit_process(plugin, context, instance=None, action=None):
         "instance": instance,
         "action": action,
         "error": None,
+        "error_info": {},
         "records": list(),
         "duration": None,
         "progress": 0,
@@ -589,8 +592,9 @@ def __implicit_process(plugin, context, instance=None, action=None):
     except Exception as error:
         lib.emit("pluginFailed", plugin=plugin, context=context,
                  instance=instance, error=error)
-        lib.extract_traceback(error)
+        error_info = lib.extract_traceback(error)
         result["error"] = error
+        result["error_info"] = error_info
 
     __end = time.time()
 
@@ -623,6 +627,7 @@ def repair(plugin, context, instance=None):
         "plugin": plugin,
         "instance": instance,
         "error": None,
+        "error_info": {},
         "records": list(),
         "duration": None
     }
@@ -643,8 +648,9 @@ def repair(plugin, context, instance=None):
             provider.invoke(plugin.repair)
             result["success"] = True
     except Exception as error:
-        lib.extract_traceback(error)
+        error_info = lib.extract_traceback(error)
         result["error"] = error
+        result["error_info"] = error_info
 
     __end = time.time()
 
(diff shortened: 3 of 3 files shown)


pyblish/cli.py
```diff
@@ -355,7 +355,7 @@ def gui(ctx, package):
 
     with _cli_plugin(data=context.data) as plugin_path:
         environ["PYBLISHPLUGINPATH"] = os.pathsep.join(
-            ctx.obj["plugin_paths"] + [plugin_path]
+            ctx.obj["plugin_paths"] + tuple(plugin_path)
```
mottosso: FYI these are not the same.

```py
>>> tuple("hello")
('h', 'e', 'l', 'l', 'o')
>>> ["hello"]
['hello']
>>>
```

Are we sure `plugin_path` is not a string?

pyblish/lib.py
```diff
@@ -57,8 +57,18 @@ def extract_traceback(exception):
     """Inject current traceback and store in exception"""
     exc_type, exc_value, exc_traceback = sys.exc_info()
     exception.traceback = traceback.extract_tb(exc_traceback)[-1]
```
mottosso, line 59: About the new `error_info` key, I think these are already part of what you can get out of the `error` key.

```py
>>> from pyblish import api
>>> class Plugin(api.ContextPlugin):
...   def process(self, context):
...     assert False, "This was a problem"
...
>>> api.register_plugin(Plugin)
>>> ctx = util.publish()
This was a problem
>>> error = ctx.data["results"][-1]["error"]
>>> error
AssertionError('This was a problem')
>>> error.traceback
<FrameSummary file <stdin>, line 3 in process>
>>> dir(error.traceback)
['__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getitem__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__iter__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__slots__', '__str__', '__subclasshook__', '_line', 'filename', 'line', 'lineno', 'locals', 'name']
>>> tb = error.traceback
>>> tb.lineno
3
>>> tb.name
'process'
>>> tb.filename
'<stdin>'
```

pyblish/cli.py
mkolar: I believe this was causing us some major headaches at one point. @illicitit might remember exactly.

pyblish/lib.py
mkolar: we'll certainly double check that maybe just missed it

pyblish/cli.py
iLLiCiTiT: if i remember well it was because `ctx.obj["plugin_paths"]` is tuple and concatenation of tuple and list is not allowed. But you got true it's not right, should be `tuple([plugin_path, ])`

pyblish/lib.py
iLLiCiTiT: not sure if it was my fault in pyblish-lite the error.traceback still was traceback object but all attributes returned `<string>` which was in GUI showed as an empty string

pyblish/cli.py
iLLiCiTiT: maybe just `(plugin_path, )`

pyblish/lib.py
iLLiCiTiT: Oh, my message disappeared... Well again:
I've investigated a little bit and found out that it was issue in Nuke and Maya:
- With pyblish-lite in Nuke or Maya `error.traceback` is not `FrameSummary` but `tuple` object. Values can be extracted with `fname, lineno, func, exc = error.traceback` but `fname` has value `"<string>"` and `exc` is `"None"` (lineno and func store values as expected)
- This issue won't happen when mocked pyblish-lite is launched -> I assume that pyblish-lite uses Nuke's or Maya's `traceback` module which has modified implementation of `extract_tb` ? (now I noticed it makes no sence because pyblish-base's method `extract_traceback` happens in same thread as pyblish-lite with the difference that there extraction works)

BTW: is possible to extract stack from error.traceback?

mottosso:
Ok, I haven't tested this myself, but based on your findings it would appear `"error"` is broken under those circumstances. In that case, rather than adding another (working) member, I would prefer it if we could "fix" the `"error"` member.

For example, could we simply override the broken members? E.g. `error.traceback.filename = str(fname)`?

iLLiCiTiT:
> For example, could we simply override the broken members? E.g. `error.traceback.filename = str(fname)`?

I will try!

iLLiCiTiT:
So the issue is that maya's module `traceback` don't have defined `FrameSummary` class -> instead of object with attributes filename, lineno,... returns tuple.
You can try run this in maya to see:
```
import sys, traceback

try:
    raise Exception("my error")
except Exception:
    exc_type, exc_value, exc_traceback = sys.exc_info()
    trc = traceback.extract_tb(exc_traceback)[-1]
    print type(trc) # result is tuple
```

That means there is not possible to use `error.traceback.filename = str(fname)` because tuple don't have attribute `filename`.
So one possibility is to use dictionary or another possibility I have in my head that won't affect base and qml is to create own class like `PyblishFrameSummary` in pyblish-base that will have attributes used in code (`filename`, `lineno`, `name`, ?`stacktrace`?) similar to traceback's FrameSummary and set that object to exception.traceback instead

mottosso:
You're right, I think what I was using above was actually Python 3.

- https://docs.python.org/3/library/traceback.html#framesummary-objects

That actually brings up an important point about Python 3, something that would make Pyblish incompatible in any host, like Maya 2020, where Python 3 was used. But that's for later.

For now, the important part is that whatever we do to change `"error"` remains compatible with how that variable is used. Have a look at what QML does with it for example (which does consider it a tuple)..

https://github.com/pyblish/pyblish-qml/blob/2d296d4b1c7239805d7e7fce6067d411fdd85c24/pyblish_qml/ipc/formatting.py#L102

From what it looks like, it's also depending on those values being accurate. Are you able to confirm whether this issue happens in QML? If you could post a reproducible with Lite, that could help narrow down the issue.

iLLiCiTiT:
> which does consider it a tuple

FrameSummary object also returns tuple if is assigned to 4 variables.

I can prepare pyblish-lite branch that will use pyblish-base master (only error.traceback) so you can test it in Maya.

But I'm really curious how pyblish-qml is launched in maya because it also seems to have same issue in another hosts:
image

iLLiCiTiT:
This is branch where pyblish-lite use error.traceback
https://github.com/pypeclub/pyblish-lite/tree/pype_with_master_base

During preparing I've remembered why I needed to use that `error_info` dictionary!
That was only way how to get stack trace which was the only source of information about where the exception happened because there is an issue in Maya's `extract_tb` which returns strange values.
```
import sys, traceback
try:
    raise Exception("my error")
except Exception:
    exc_type, exc_value, exc_traceback = sys.exc_info()
    fname, line_no, func, exc = traceback.extract_tb(exc_traceback)[-1]

    print fname # in Maya console returns <maya console> in pyblish it is <string>
    print line_no # is actually lineno+1 ?
    print func
    print exc # is not defined
```

With these information I expect that pyblish-qml is launched in Maya with subprocessed Python so traceback module works as he should?

EDIT:
I tried Pyblish QML in Maya with same result. All the time I though that pyblish qml show errors in Maya with all values.
(I'm using Maya 2018 for testing)

BigRoy:
I noticed this PR has been open for a while. What's the current status and what exactly does still need to change or be tested to make sure we can get to finalizing it?

mottosso:
It's in a no-go state, at the moment. Because it adds one missing thing, but then duplicates others.

What we could and should add is https://github.com/pyblish/pyblish-base/pull/347/files#diff-92b154b8082f83fc98dc387160ecea9eR61, which is the stacktrace itself. I would nuke this PR, and make a new one with just that. One that adds to the existing `context.data["error"]` as `.stacktrace`, rather than creating a new dictionary "error_info".

PaulSchweizer:
Hi @mottosso just wanted to check in on the state of this. The mentioned addition of having the formated exception available would be of great use to me! Let me know if I can help in any way.

mottosso:
It's ready for anyone in need of it put a PR together. 👍

PaulSchweizer:
I submitted a PR: https://github.com/pyblish/pyblish-base/pull/356

Issue #348, opened 2019-09-01
Pyblish profiling


antirotor:
### Goal

It would be nice to have mode, that would generate profiling data for pyblish process and plugins. Producing lot of timing and other data, that can be later used for visualizing and debugging whole session.

### Motivation

My idea is to capture whole context after each plugin is done as snapshots along with it's timing, memory consumtion, etc. to json. Then to have simple application processing this json file, drawing all data to canvas, so we'll be able to vizualize, what is going on with all data during publishing. This could be useful when debugging behaviour of pipelines where lot of plugins are doing many things.

### Implementation

We could use similar way already used all over pyblish - use registered callbacks to be run after each plugin and use python's [cProfiler](https://docs.python.org/3/library/profile.html#module-cProfile) or other stuff to measure.

mottosso:
Thanks @antirotor this sounds like a good and relatively easy thing to implement.

```py
import json
from pyblish import util
context = util.publish()

with open("profile.json", "w") as f:
  json.dump(f, context.data["profiling"])
```

I would have a look at building this into the main [`process()`](https://github.com/pyblish/pyblish-base/blob/f8a2449af39199c2620c3d402d28ed21a6ae974d/pyblish/plugin.py#L451) function; this is what standalone and GUIs calls to actually process a plug-in/instance pair. There could be e.g. a start/end time before and after that function finishes.

The data itself could be stored in e.g. `context.data["profiling"]`.

Then we could potentially also have some means of deep profiling by the plug-in author calling some method from within the plug-in, that could then be collected alongside the global time for the entire plug-in.

```py
class MyPlugin(api.InstancePlugin):
  def process(self, instance):
    # do something heavy
    self.profile("somethingHeavy", time.time())
    # do something else
    self.profile("somethingElse", time.time())
```

The `profile()` function would then add a child entry to the overally plug-in time. It'd need a more suitable name than `self.profile()` and probably wouldn't need to take in a time, but you get the idea.

For memory it would probably be sufficient to look at the overall consumption of the Python process itself; since it's unlikely that anything else happens during publishing. At least for an overall take on it. Then there is potentially room to monitor the context and instance objects themselves, and measure their size over time. Speaking of which, you probably also want to store clock-time alongside each process() and "deep" profile call, to build some sort of graph (with e.g. pygal) afterwards. That's something that could potentially be shipped alongside Pyblish, to generate SVG graphcs out-of-the-box. Something that could potentially also be included in the GUIs too.

[...]

PR #349, opened 2019-09-02
basic profiling support

antirotor:
### Goal
It would be nice to have mode, that would generate profiling data for pyblish process and plugins. Producing lot of timing and other data, that can be later used for visualizing and debugging whole session.

### Motivation
My idea is to capture whole context after each plugin is done as snapshots along with it's timing, memory consumtion, etc. to json. Then to have simple application processing this json file, drawing all data to canvas, so we'll be able to vizualize, what is going on with all data during publishing. This could be useful when debugging behaviour of pipelines where lot of plugins are doing many things.

This implements #348

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -63,35 +63,40 @@ STRICT_DATATYPES = EARLY_ADOPTER or STRICT_DATATYPES
 class ProfilingTimer():
     """Providing context manager timing and memory usage
 
-       TODO: find a way to do memory profiling on windows without additional
-       external dependency.
+    If `psutil` module is found, it will add memory consumption information
+    to profile data. You can install it with `pip install psutil` - on linux
+    it might require python headers and dev tools (on Centos for example).
     """
 
     def __init__(self):
         pass
 
     def __enter__(self):
-        # resource module is supported only on POSIX platforms
-        if sys.platform.startswith('linux') or sys.platform == 'darwin':
-            import resource
-            self.initial_shared_memory = resource.getrusage(
-                resource.RUSAGE_SELF).ru_ixrss
-            self.initial_private_memory = resource.getrusage(
-                resource.RUSAGE_SELF).ru_idrss
+        try:
+            import psutil
+        except ImportError:
+            # psutil not found, memory information will be missing
+            pass
+        else:
+            process = psutil.Process(os.getpid())
+            # in MB
+            self.initial_mem = process.get_memory_info()[0] / float(2 ** 20)
+
         self.start = time.time()
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.end = time.time()
         self.runtime = self.end - self.start
-        if sys.platform.startswith('linux') or sys.platform == 'darwin':
-            import resource
-            self.final_shared_memory = resource.getrusage(
-                resource.RUSAGE_SELF).ru_ixrss
-            self.final_private_memory = resource.getrusage(
-                resource.RUSAGE_SELF).ru_idrss
-            self.max_memory = resource.getrusage(
-                resource.RUSAGE_SELF).ru_maxrss
+        try:
+            import psutil
+        except ImportError:
+            # psutil not found, memory information will be missing
+            pass
+        else:
+            process = psutil.Process(os.getpid())
+            # in MB
+            self.final_mem = process.get_memory_info()[0] / float(2 ** 20)
 
 
 class Provider():
@@ -512,14 +517,13 @@ def process(plugin, context, instance=None, action=None):
         "runtime": profiler.runtime,
     }
 
-    # memory usage data aren't easily available on windows platform
-    if sys.platform.startswith('linux') or sys.platform == 'darwin':
-        profiling_data['max_memory'] = profiler.max_memory
-        shared_memory = profiler.final_shared_memory - profiler.initial_shared_memory  # noqa: E501
-        profiling_data['shared_memory_change'] = shared_memory
-        private_memory = profiler.final_private_memory - profiler.initial_private_memory  # noqa: E501
-        profiling_data['private_memory_change'] = private_memory
-        profiling_data['max_memory_usage'] = profiler.max_memory
+    # collected using psutil module, None if it is not installed
+    try:
+        profiling_data['initial_memory_sage'] = profiler.initial_mem
+        profiling_data['final_memory_usage'] = profiler.final_mem
+    except AttributeError:
+        # no memory information collected
+        pass
 
     context.data['profiling'][plugin] = profiling_data
     profiling_data["snapshot"] = {
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -20,6 +20,8 @@ import inspect
 import warnings
 import contextlib
 import uuid
+import json
+import tempfile
 
 # Local library
 from . import (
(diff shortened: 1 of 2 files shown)


antirotor:
This still have some caveats. Getting memory information on windows is somehow cumbersome without additional external dependencies (such as [psutils](https://pypi.org/project/psutil). Problem with **psutils** is that it needs developer version of python on linux (gcc and python header), complicating installation somehow.

Also missing tool to process profiling info yet.

mottosso:
Thanks for this.

Re psutil; it's kind of the de-facto standard for doing these kinds of OS queries; how about we depend on it as an optional dependency? That is, if it's available, the profiling will be that more thorough, and without it you get the basics. That way, users on Windows can `pip install psutil` and those on Linux could e.g. `apt-get install python-psutil` or whatever means are available, if they are interested in that kind of information.

If there is a pure-Python equivalent, that would of course be best. But in my experience it just isn't possible. :(

antirotor:
I've changed it to use optional **psutil** module. Pure python isn't there - only **resource** module but that is available only on linux and darwin systems.

pyblish/plugin.py
```diff
@@ -462,10 +503,38 @@ def process(plugin, context, instance=None, action=None):
...
+    context.data['profiling'][plugin] = profiling_data
+    profiling_data["snapshot"] = {
+        "timestamp": int(time.time()),
+        "context": json.dumps(context, default=lambda o: '<not serializable>')
+    }
+
+    profiling_path = os.path.join(tempfile.gettempdir(),
+                                  "{}-pyblish_profile.json".format(context.id))
+
+    append_to_json(profiling_data, profiling_path)
```
mottosso, line 537: Ah no, I don't think this is a good idea.

Why not let the user decide what to do with the profiling information?

```py
context = util.publish()

with open("somefile.json", "w") as f:
  json.dump(f, context.data["profiling"])
```

GUIs could get a hold of it via the context as well, inbetween each process if need be, but likely also as a post-process of sorts.

pyblish/plugin.py
```diff
@@ -462,10 +503,38 @@ def process(plugin, context, instance=None, action=None):
...
+        else:
+            result = __implicit_process(plugin, context, instance, action)
+
+    profiling_data = {
+        "runtime": profiler.runtime,
+    }
+
+    # collected using psutil module, None if it is not installed
+    try:
+        profiling_data['initial_memory_sage'] = profiler.initial_mem
```
mottosso, line 522: Typo on `_sage`. The naming convention should be `mixedCase` for dictionary keys as well (JSON convention).

pyblish/plugin.py
```diff
@@ -462,10 +503,38 @@ def process(plugin, context, instance=None, action=None):
 
     """
 
-    if issubclass(plugin, (ContextPlugin, InstancePlugin)):
-        result = __explicit_process(plugin, context, instance, action)
-    else:
-        result = __implicit_process(plugin, context, instance, action)
+    if not context.get('profiling'):
```
mottosso, line 506: Typo here, should be `context.data.get()`, as below.

Unsure about whether `profiling` is the best name here. Grammatically it's a verb, describing some sort of state. That it is currently "profiling". As it isn't meant as that, it should probably we `profile`, but even that could be confused with an action. To "profile" something.

I'm thinking `context.data["statistics"]`

pyblish/plugin.py
```diff
@@ -58,6 +60,45 @@
 STRICT_DATATYPES = EARLY_ADOPTER or STRICT_DATATYPES
 
 
+class ProfilingTimer():
```
mottosso, line 63: I think we should stick with a `@contextmanager`, like it's being used elsewhere in Pyblish.

```python
import contextlib

@contextlib.contextmanager
def profiling():
  stats = {}
  stats["startTime"] = time.time()
  yield stats
  stats["endTime"] = time.time()

with profiling() as stats:
  ...

```

Alternatively..

```py
stats = {}
with profiling(stats):
  ...
```

pyblish/plugin.py
```diff
@@ -462,10 +503,38 @@ def process(plugin, context, instance=None, action=None):
...
+        profiling_data['initial_memory_sage'] = profiler.initial_mem
+        profiling_data['final_memory_usage'] = profiler.final_mem
+    except AttributeError:
+        # no memory information collected
+        pass
+
+    context.data['profiling'][plugin] = profiling_data
+    profiling_data["snapshot"] = {
+        "timestamp": int(time.time()),
+        "context": json.dumps(context, default=lambda o: '<not serializable>')
```
mottosso, line 531: Not sure about "snapshot".. If the goal is capturing more data, I think we should instead favour a mechanism with which to let the plug-in author provide his own data, as mentioned in https://github.com/pyblish/pyblish-base/issues/348#issuecomment-527023863

pyblish/plugin.py
```diff
@@ -462,10 +503,38 @@ def process(plugin, context, instance=None, action=None):
...
+    try:
+        profiling_data['initial_memory_sage'] = profiler.initial_mem
+        profiling_data['final_memory_usage'] = profiler.final_mem
+    except AttributeError:
+        # no memory information collected
+        pass
+
+    context.data['profiling'][plugin] = profiling_data
+    profiling_data["snapshot"] = {
+        "timestamp": int(time.time()),
```
mottosso, line 530: If we keep `startTime` from `stats` we can remove this.

mottosso (commented):
Sorry for the delay in looking at this, I've made some notes in-line, and overall I think there's one thing missing, which is that profiling currently happens per-process, but are stored as one global object, being overwritten each time.

Instead, I think we should store them together with the `context.data["results"]` data. And actually, looking through the code, I found that we're already collecting the duration of each process in the results dict.

https://github.com/pyblish/pyblish-base/blob/f8a2449af39199c2620c3d402d28ed21a6ae974d/pyblish/plugin.py#L532-L534

This would be a good spot to continue adding profile data I think, so let's move the profiling context manager to there.

Let me know what you think.

pyblish/plugin.py
antirotor: So you think that it shouldn't save to json file implicitly and it's up to user to handle data in `context.data.get('statistics')` (lets call it that) afterward? Just checking if I understand it correctly.

pyblish/plugin.py
antirotor: I agree, *statistics* is much better.

pyblish/plugin.py
antirotor: I was thinking about snapshot as "map" if you will, that with all those profiling data you could later on visualize whole pyblish process with all its timings and what was changed in context each step. But yes, adding custom data from every plugin should be implemented also.

pyblish/plugin.py
mottosso: Yes, don't write files without the user's knowledge. In the vast majority of cases, the user won't care for those files anyway. Profiling is primarily useful for the plug-in author.

mottosso:
Hi @antirotor, revisiting this long-forgotten issue and would like to merge it. It's a good feature! There are just a few notes left before a first 1.0 of the feature, if you're up for it it would be good to see those though.

antirotor:
> Hi @antirotor, revisiting this long-forgotten issue and would like to merge it. It's a good feature! There are just a few notes left before a first 1.0 of the feature, if you're up for it it would be good to see those though.

Sure thing, sorry for this zombie-state. I had more urgent things to do but it is still on my mind. I'll try to take care of it asap.

PR #356, opened 2019-11-04
Traceback available in error results

PaulSchweizer:
Based on this PR/discussion: https://github.com/pyblish/pyblish-base/pull/347#issuecomment-548791773

I also added logging of the error, useful for situations where there is no UI and/or something unexpected occurs but I'm not sure if there are arguments against this?

As for the unittest, I am not sure if this is the correct location, please let me know where the test should go.

### DIFF
--- a/pyblish/lib.py
+++ b/pyblish/lib.py
@@ -57,7 +57,17 @@ def extract_traceback(exception):
     """Inject current traceback and store in exception"""
     exc_type, exc_value, exc_traceback = sys.exc_info()
     exception.traceback = traceback.extract_tb(exc_traceback)[-1]
+    filename, lineno, func, exc = exception.traceback
+    error_info = {
+        'msg': str(exception),
+        'filename': str(filename),
+        'lineno': str(lineno),
+        'func': str(func),
+        'traceback': ''.join(traceback.format_exception(
+            exc_type, exc_value, exc_traceback))
+    }
     del(exc_type, exc_value, exc_traceback)
+    return error_info
 
 
 def time():
--- a/pyblish/lib.py
+++ b/pyblish/lib.py
@@ -53,18 +53,33 @@ class MessageHandler(logging.Handler):
             self.records.append(record)
 
 
-def extract_traceback(exception):
-    """Inject current traceback and store in exception"""
+def extract_traceback(exception, plugin):
+    """Inject current traceback and store in exception.exception
+
+    Arguments:
+        exception (Exception): Exception object
+        plugin (pyblish.api.Plugin): Plugin class that caused the exception.
+            This is necessary to inject the correct file path in the traceback.
+            If plugins are registered through `api.plugin.discover`, they only
+            show "<string>" instead of the actual source file.
+    """
     exc_type, exc_value, exc_traceback = sys.exc_info()
     exception.traceback = traceback.extract_tb(exc_traceback)[-1]
     filename, lineno, func, exc = exception.traceback
+
+    tb = ''.join(traceback.format_exception(
+        exc_type, exc_value, exc_traceback))
+    if 'File "<string>", line ' in tb:
+        filename = plugin.__module__
+        tb = tb.replace(
+            'File "<string>", line ',
+            'File "{0}", line'.format(plugin.__module__))
     error_info = {
         'msg': str(exception),
         'filename': str(filename),
         'lineno': str(lineno),
         'func': str(func),
-        'traceback': ''.join(traceback.format_exception(
-            exc_type, exc_value, exc_traceback))
+        'traceback': tb
     }
     del(exc_type, exc_value, exc_traceback)
     return error_info
--- a/pyblish/lib.py
+++ b/pyblish/lib.py
@@ -54,7 +54,9 @@ class MessageHandler(logging.Handler):
 
 
 def extract_traceback(exception, plugin):
-    """Inject current traceback and store in exception.exception
+    """Inject current traceback and store in exception.traceback.
+
+    Also storing the formatted traceback on exception.formtatted_traceback.
 
     Arguments:
         exception (Exception): Exception object
@@ -65,24 +67,16 @@ def extract_traceback(exception, plugin):
     """
     exc_type, exc_value, exc_traceback = sys.exc_info()
     exception.traceback = traceback.extract_tb(exc_traceback)[-1]
-    filename, lineno, func, exc = exception.traceback
 
-    tb = ''.join(traceback.format_exception(
+    formatted_traceback = ''.join(traceback.format_exception(
         exc_type, exc_value, exc_traceback))
-    if 'File "<string>", line ' in tb:
-        filename = plugin.__module__
-        tb = tb.replace(
+    if 'File "<string>", line ' in formatted_traceback:
+        formatted_traceback = formatted_traceback.replace(
             'File "<string>", line ',
             'File "{0}", line'.format(plugin.__module__))
-    error_info = {
-        'msg': str(exception),
-        'filename': str(filename),
-        'lineno': str(lineno),
-        'func': str(func),
-        'traceback': tb
-    }
+    exception.formatted_traceback = formatted_traceback
+
     del(exc_type, exc_value, exc_traceback)
-    return error_info
 
 
 def time():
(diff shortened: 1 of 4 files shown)


tests/plugins.py
```diff
@@ -0,0 +1,24 @@
+"""Plugins for testing purposes.
```
PaulSchweizer, line 1: Added this file so the test can use the default discover behaviour, leading to the issue explained above (File "<string>" instead of the actual file name). Having these tests in the test file itself will not trigger that issue.

tests/plugins.py
```diff
@@ -0,0 +1,24 @@
...
+
+Soruce them like this from within a test function:
```
BigRoy: Typo: `Soruce` -> `source`

tests/test_logic.py
```diff
@@ -185,3 +185,47 @@ class ClassF(api.Collector):
 
     assert logic.plugins_by_families(
         [ClassD, ClassE, ClassF], ["a", "b", "c"]) == [ClassD, ClassE]
+
+
+REGISTERED_PATHS = []
```
mottosso: You can replace this with the decorator `@with_setup(lib.setup_empty, lib.teardown)`, which will empty any trace of a plug-in before calling the test, and restore things on exit. Much like what you've done here. See how it's used elsewhere to get an idea of how to use it.

tests/plugins.py
mottosso: Ok, that looks fine.

mottosso:
Ok, I've had a look at this and unfortunately it reproduces the problem with #347, in that we now have duplicate data.

Here's the data present in the `context.data["results"][-1]["error"]` member.

**test.py**

```py
from pyblish import api, util

class Plugin(api.ContextPlugin):
   def process(self, context):
     assert False, "This was a problem"

api.register_plugin(Plugin)
ctx = util.publish()
error = ctx.data["results"][-1]["error"]

fname, lineno, func, msg = error.traceback
print(fname, lineno, func, msg)
```

**Python 2**

```bash
> $ python .\test.py
This was a problem
('.\\test.py', 5, 'process', 'assert False, "This was a problem"')
```

**Python 3**

```bash
> $ python .\test.py
This was a problem
.\test.py 5 process assert False, "This was a problem"
```

Notice how all information in the new `error_info` is all here, unless I'm missing something?

In addition, this PR includes support for a `<string>` filename, which can (only?) happen when registering a plug-in in an interactive Python session. And outside of experimentation and testing, is that something people do? For example, writing in the Maya Script Editor and using that in their production publishes, or somehow reading the raw test from a Python module of Pyblish plug-ins, and `exec()`-ing those.

We could support that, like you have here, but we would need a usecase for it. A reason.

Other than supporting `<string>`, does this PR add anything new? Could it be the case that `"error"` was misunderstood, or not visible enough? Granted, it isn't obvious that the members contained in that tuple are in that order, but that's what Python's [traceback object looks like](https://docs.python.org/2/library/traceback.html#traceback.extract_tb).

PaulSchweizer:
**Duplicated data**
Yes kind of. I added the error_info dict from that PR because I was under the false impression that that was what was agreed upon, I do not need it.
The only thing we do not yet have (and I would want) is the formated traceback message which I can not reproduce from `fname, lineno, func, msg = error.traceback` after the fact since the `sys.exc_info()` is gone by that time. If there is a way to achieve that though, please let me know.

Basically this is the only thing we'd need:

**test.py**

```python
from pyblish import api, util

class Plugin(api.ContextPlugin):
   def process(self, context):
     assert False, "This was a problem"

api.register_plugin(Plugin)
ctx = util.publish()
error = ctx.data["results"][-1]["error"]

fname, lineno, func, msg = error.traceback
formatted_traceback = error.formatted_traceback  # This would be new
print(fname, lineno, func, msg)
print(formatted_traceback)
```

**Result:**

```
> $ python .\test.py
This was a problem
('.\\test.py', 5, 'process', 'assert False, "This was a problem"')

  File "/pyblish/plugin.py", line 518, in __explicit_process
    runner(*args)
  File "/test.py", line 5, in process
AssertionError: This was a problem 
```  

**String paths**
We are registering pyblish plugins like this everywhere:

```python
from pyblish import api 

api.register_plugin_path("my/plugin/path")
api.discover()
```

Which leads to the `<string>` issue as reproduced in the tests but please let me know if this does not happen for you?
Also, let me know if we should refrain from doing it this way? It's just helpful in situations where we have plugins in context specific places, but that can be solved differently as well of course.

BigRoy:
👍 To add a little note from my end.

Exactly this is what I'd like to see too:
```python
  File "/pyblish/plugin.py", line 518, in __explicit_process
    runner(*args)
  File "/test.py", line 5, in process
AssertionError: This was a problem 
```

Especially the full traceback that also includes this information: `File "/pyblish/plugin.py", line 518, in __explicit_process`. Note how that information is not in the other data. This would allow to have a full traceback depth as opposed to only the last call. If I recall correctly that's exactly the issue I was having with displayed errors in Pyblish QML, that they were incapable of displaying the full traceback. But I'd have to test whether that's actually (still) the case.

mottosso:
> Basically this is the only thing we'd need:

Yes, that looks good to me.

> Also, let me know if we should refrain from doing it this way?

No that's the right way, and I see the problem now. This is a bug. Plug-ins coming from `register_plugin_path` *does* have a path. But it looks like that path is thrown away by `discover()` (to facilitate the auto-reload functionality).

I had a look to see whether there was an easy fix here, but came up short. The problem is how `extract_traceback` finds that path to begin with. The discover function *does* make a record of what the path is, but it doesn't seem to be looking there.

So in addition to your fix for converting `<string>` to a full path, could you also add this?

```python
        lib.extract_traceback(error)
        if error.traceback[0] == "<string>":
           error.traceback = (os.path.abspath(plugin.__module__),) + error.traceback[1:]
        result["error"] = error
```

Idealy, the formatting would then be able to use this path, so we wouldn't have to check for `<string>` twice, but if not then that's fine.

With this, we should have a fully correct error.

PaulSchweizer:
> So in addition to your fix for converting <string> to a full path, could you also add this?

Done

> Idealy, the formatting would then be able to use this path, so we wouldn't have to check for <string> twice, but if not then that's fine.

The formatting of the exception message is an internal process of the traceback module and it relies on sys.exc_info and not the extracted data so I don't see a way to inject the correct file path in there. 
Good news though is that we only need to check for <string> once anyways.

pyblish/lib.py
```diff
@@ -53,10 +53,31 @@ def emit(self, record):
...
+            show "<string>" instead of the actual source file.
+    """
     exc_type, exc_value, exc_traceback = sys.exc_info()
     exception.traceback = traceback.extract_tb(exc_traceback)[-1]
+
+    formatted_traceback = ''.join(traceback.format_exception(
+        exc_type, exc_value, exc_traceback))
+    if 'File "<string>", line ' in formatted_traceback:
+        fname, lineno, func, msg = exception.traceback
+        exception.traceback = (plugin.__module__, lineno, func, msg)
```
mottosso: Don't forget `os.path.abspath(plugin.__module__)` to account for when paths are relative your working directory.

E.g.

1. `cd c:\some\path`
2. `publish()`, for which plugin is at `c:\some\path\plugins\my_plugin.py`
3. Path then becomes `plugins\my_plugin.py`

PaulSchweizer: Good call, thanks, it's fixed

pyblish/lib.py
```diff
@@ -53,10 +53,32 @@ def emit(self, record):
             self.records.append(record)
 
 
-def extract_traceback(exception):
-    """Inject current traceback and store in exception"""
+def extract_traceback(exception, plugin):
```
mottosso: Sorry to keep picking on this, but something only become apparent when you're just about to merge. xD

Could we make this function take a `fname=None` argument instead? Having it take `plugin` implies some dependency on Pyblish's Plugin class, and it's actually quite general and has to do with overriding the `fname` found in the `exception` argument. That would make it squeaky clean.

If you're done with this, I understand and could take a look at it myself in the coming days.

PaulSchweizer: No worries, I'll add that later today

PaulSchweizer: It's changed

mottosso:
Thanks, great work!

tokejepsen:
Awesome stuff @PaulSchweizer !

mkolar:
I completely missed this happening. 

Yeah, this will work for us as well in our pyblish lite. The result should be exactly what we were after in #347. 

Nice job.

PR #357, opened 2019-11-26
convert tuple ctx.obj["plugin_paths"] to list before concatenation

iLLiCiTiT:
- ctx.obj["plugin_paths"] may be tuple which cause issues when trying to concatenate with list
- this fix will always convert the ctx.obj["plugin_paths"] to list to avoid this issue

### DIFF
--- a/pyblish/cli.py
+++ b/pyblish/cli.py
@@ -198,6 +198,11 @@ def main(ctx,
     global _ctx
     _ctx = ctx
 
+    # Convert multi-arguments from tuple to list, see #357
+    plugin_paths = list(plugin_paths)
+    add_plugin_paths = list(add_plugin_paths)
+    data = list(data)
+
     level = LOG_LEVEL[logging_level]
     log.setLevel(level)
 
@@ -309,6 +314,10 @@ def publish(ctx,
 
     """
 
+    # Convert multi-arguments from tuple to list, see #357
+    targets = list(targets)
+    instances = list(instances)
+
     _start = time.time()  # Benchmark
 
     # Use `path` argument as initial data for context
@@ -355,7 +364,7 @@ def gui(ctx, package):
 
     with _cli_plugin(data=context.data) as plugin_path:
         environ["PYBLISHPLUGINPATH"] = os.pathsep.join(
-            list(ctx.obj["plugin_paths"]) + [plugin_path]
+            ctx.obj["plugin_paths"] + [plugin_path]
         )
 
         process = subprocess.Popen(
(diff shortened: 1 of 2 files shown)


tokejepsen:
When does this return a tuple?

Maybe we can get a test for this, to cement it for the future?

iLLiCiTiT:
We actually didn't found out when/why it happens. Seems like it happens when pyblish is launched with cli in python 3 (but not sure).  We're using these args: `-pp {paths_to_plugins}` and `-d {data_info}`

EDITED:
and `--publish-gui` argument

iLLiCiTiT:
Is possible to merge this?

davidlatwe:
It seems that module `click` which responsible to parsing args was meant to parse *multiple options* as `tuple`, see [here](https://click.palletsprojects.com/en/3.x/options/#multiple-options).

In doc it says *Multiple Options* is similarly to `nargs`, which configures [*Multi Value Options*](https://click.palletsprojects.com/en/3.x/options/#multi-value-options), and it stores values as a tuple.

Look at the code in `.vendor.click.core` in this [line](https://github.com/pyblish/pyblish-base/blob/e58e21e44edd7b78ce814788c43161ff5f0f465d/pyblish/vendor/click/core.py#L958), if you replace `tuple` with `list`, the error we has encountered is gone.

pyblish/cli.py
```diff
@@ -355,7 +355,7 @@ def gui(ctx, package):
 
     with _cli_plugin(data=context.data) as plugin_path:
         environ["PYBLISHPLUGINPATH"] = os.pathsep.join(
-            ctx.obj["plugin_paths"] + [plugin_path]
+            list(ctx.obj["plugin_paths"]) + [plugin_path]
```
davidlatwe: IMHO, maybe it's better to do this conversion at the beginning of the function, like
```python
def gui(ctx, package):

    environ = os.environ.copy()
    context = ctx.obj["context"]
    plugin_paths = list(ctx.obj["plugin_paths"])
    ...
```

mottosso: I think that's a good idea, but I think we can go even further, and convert those as soon as they come out of Click. Having a look now.

mottosso:
There we go. Once someone can confirm this solves the issue this is ready to merge. Now values are list regardless of whether they come from Pyblish or CLI.

iLLiCiTiT:
I checked and `TypeError: can only concatenate tuple (not "list") to tuple` did not appear so I assume it works.

davidlatwe:
Nice ! Can confirm it works !

PR #358, opened 2019-12-12
Iterator now sorting the plugins before iteration

PaulSchweizer:
Also updated the respective test and added venv to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ build
 playground.py
 *.sublime-*
 __pycache__
+venv/
--- a/pyblish/logic.py
+++ b/pyblish/logic.py
@@ -14,7 +14,9 @@ from .plugin import (
     Subset,
     Exact,
 
-    registered_targets
+    registered_targets,
+
+    sort
 )
 
 _algorithms = {
@@ -358,7 +360,7 @@ def Iterator(plugins, context, state=None):
     # when running the Iterator directly without registering any targets.
     targets = registered_targets() or ["default"]
 
-    plugins = plugins_by_targets(plugins, targets)
+    plugins = sort(plugins_by_targets(plugins, targets))
 
     for plugin in plugins:
         if not plugin.active:
--- a/tests/test_logic.py
+++ b/tests/test_logic.py
@@ -23,7 +23,10 @@ def no_guis():
 
 @with_setup(lib.setup, lib.teardown)
 def test_iterator():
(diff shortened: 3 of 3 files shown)


tokejepsen:
Cool!

What does this solve? Is there an issue to reference to this?

PaulSchweizer:
I came across this when using the `pyblish.util.integrate` function (as well as the functions for the other steps) and thought this should be fixed as it does not yield the same behaviour as the "regular" pyblish method where the plugins are pre-sorted.
Should I open a ticket? Or use things in a different way?

tokejepsen:
The best way to would be make a test, so we can ensure functionality in the future as well.

PaulSchweizer:
The test is are already in place, as mentioned above

mottosso:
This affects a core mechanic of Pyblish, and thus makes me nervous. :S

Plug-ins are sorted [during discovery](https://github.com/pyblish/pyblish-base/blob/b3e215740a4224e1a04c7fd77c3fb0a497ec6fbf/pyblish/plugin.py#L1375), I can't figure out why they'd need to be sorted again afterwards?

Maybe if you could post a reprodible?

PaulSchweizer:
Maybe I'm using it wrong, but I am providing a list of plugins to the collect, validate etc. functions https://github.com/pyblish/pyblish-base/blob/master/pyblish/util.py#L199 

That list might not be in order, in which case it is exectuted out of order (meaning in it's given order).

I am using these functions separately for two reasons:
- We have a standalone tool that only performs validation on a given set of selectable validators
- We have our own ui for the entire publish process and it has to use these functions as well.

I can of course sort the list of plugins beforehand myself, but thought it safer if the sorting is performed inside of pyblish to avoid this inconsistency.

Of yourse there is also the possibility that the mentioned functions are not meant to be used like I am using them.

In the tests I added the above mentioned case, providing a list of plugins in the wrong order: https://github.com/pyblish/pyblish-base/pull/358/files#diff-08aca4e83f1cbb3f29db48aa9abcae9eR65

If that is not sufficient I can draft up a lengthier example, just let me know

mottosso:
> but I am providing a list of plugins to the collect, validate etc.

Ah, yes. You should be able to pass a list of your order. If you wanted it sorted, then you can sort it before handing it in. This is expected, as you may want to actually control that order.

PaulSchweizer:
Ok I see, then this PR is void, thanks for clarifying!

PR #362, opened 2020-03-02
Traceback `line` without space

iLLiCiTiT:
Cosmetic change:
- added missing space to replacement string in traceback formatting


mottosso:
Nitpick, but would it maybe make more sense to *remove* the additional space from the source string?

iLLiCiTiT:
Now when you talk about it, I think so too :) will change

mottosso:
Looks good to me, thanks. Make a new PR with a version increment if you want it released immediately.

Issue #345, opened 2019-07-15
Explicitly providing targets to util.publish() or pyblish_qml.show() excludes "default" target


BigRoy:
### Issue

In Pyblish whenever one registers a custom target it still includes the `default` target - as if it is **always** present.

```python
import pyblish.api
import pyblish.util
pyblish.api.register_target("render")

pyblish.util.publish()
# Pyblish will now use targets: ["default", "render"]
```

This concept propagates correctly to `pyblish_qml.show()` when *no* targets are provided:

```python
import pyblish.api
import pyblish_qml
pyblish.api.register_target("render")

pyblish_qml.show()
# Pyblish QML will now use targets: ["default", "render"]
```

However when providing targets explicitly to `pyblish_qml.show()` it will only include `default` when explicitly provided. This behavior does nicely match what Pyblish currently does when using `pyblish.util.publish()`, like:

```python
pyblish.util.publish(targets=["render"])
# Pyblish will only use targets: ["render"]

pyblish_qml.show(targets=["render"])
# Pyblish QML will only use targets: ["render"]
```
_Note that here "default" is **not** included in the targets_

---

### Discussion

The consusion rises when you query `pyblish.api.registered_targets()` , it never lists "default" even though it is included by default, yet not when you start to explicitly providing targets.

Somehow it left me confused for two hours where the behavior started to differ when I triggered QML with a target and it lacked the default plug-ins. I'm wondering whether we should be able to simplify the logic so it's the same in all those cases. Additionally the [*Documentation* led me off the path too as it indicated by default plug-ins would target anything with `*`](https://api.pyblish.com/pages/targets.html):

> Targets work the same way as families, so a wildcard of `*` enables the plugin for all targets. Since all plugins are registered with `*` as targets, this workflow is backwards compatible with existing plugins that has not overwritten the targets attribute.

However the [code indicates the default target is set only to `targets = ["default"]`](https://github.com/pyblish/pyblish-base/blob/f8a2449af39199c2620c3d402d28ed21a6ae974d/pyblish/plugin.py#L241)

---

Reference:
- [Discussion on Pyblish gitter about targets behavior](https://gitter.im/pyblish/pyblish?at=5d2c865b4d750532414d24b1)

BigRoy:
[...]

PR #364, opened 2020-03-09
Explicit targets not registered globally

### LINKED ISSUE #363: Explicitly use targets without global registration
The [pyblish.util.publish()](https://github.com/pyblish/pyblish-base/blob/acf97ef9fdc2fc506daac7817db99073a79a7c56/pyblish/util.py#L30) function (and all helpers for specific CVEI stage) can take a list of targets but [pyblish.util._convenience_iter()](https://github.com/pyblish/pyblish-base/blob/acf97ef9fdc2fc506daac7817db99073a79a7c56/pyblish/util.py#L96) uses the global registration in the backend, which causes a bunch of issues:

* Targets previously registered get de-registered when process ends (as described in #345)

```python
>>> pyblish.plugin.register_target("foo")
>>> print(pyblish.api.registered_targets())
["foo"]
>>> pyblish.util.publish(targets=["foo", "bar"])
>>> print(pyblish.api.registered_targets())
[]
```

* It is impossible to run 2 processes simultaneously with different targets

```python
>>> t1 = threading.Thread(target=pyblish.util.publish, kwargs={"targets": ["foo"]})
>>> t2 = threading.Thread(target=pyblish.util.publish, kwargs={"targets": ["bar"]})
>>> t1.start() # Register "foo"
>>> t2.start() # Register "bar" while t1 is still running...
```

It would be nice to modify the logic so that targets explicitly passed to these functions don't get registered globally. What do you think?

buddly27:
This resolves #363 and #345

### DIFF
--- a/pyblish/logic.py
+++ b/pyblish/logic.py
@@ -356,10 +356,12 @@ def Iterator(plugins, context, state=None, targets=None):
     }
 
     # Include "default" target when no targets are requested.
-    targets = targets or ["default"]
-    targets += registered_targets()
+    _targets = targets or ["default"]
 
-    plugins = plugins_by_targets(plugins, targets)
+    if targets is None:
+        _targets += registered_targets()
+
+    plugins = plugins_by_targets(plugins, _targets)
 
     for plugin in plugins:
         if not plugin.active:
--- a/pyblish/logic.py
+++ b/pyblish/logic.py
@@ -334,7 +334,7 @@ def _extract_traceback(exception):
         del(exc_type, exc_value, exc_traceback)
 
 
-def Iterator(plugins, context, state=None):
+def Iterator(plugins, context, state=None, targets=None):
     """Primary iterator
 
     This is the brains of publishing. It handles logic related
@@ -345,6 +345,7 @@ def Iterator(plugins, context, state=None):
         plugins (list): Plug-ins to consider
         context (list): Instances to consider
         state (dict): Mutable state
+        targets (list, optional): Targets to include for publish session.
 
     """
 
@@ -356,7 +357,7 @@ def Iterator(plugins, context, state=None):
 
     # We'll add "default" target if no targets are registered. This happens
     # when running the Iterator directly without registering any targets.
-    targets = registered_targets() or ["default"]
+    targets = targets or registered_targets() or ["default"]
 
     plugins = plugins_by_targets(plugins, targets)
 
--- a/pyblish/logic.py
+++ b/pyblish/logic.py
@@ -355,13 +355,12 @@ def Iterator(plugins, context, state=None, targets=None):
         "ordersWithError": set()
     }
 
-    # Include "default" target when no targets are requested.
-    _targets = targets or ["default"]
+    # Include "default" target and registered targets when no targets are
+    # explicitly requested.
+    if not targets:
+        targets = ["default"] + registered_targets()
 
-    if targets is None:
-        _targets += registered_targets()
-
-    plugins = plugins_by_targets(plugins, _targets)
+    plugins = plugins_by_targets(plugins, targets)
 
     for plugin in plugins:
         if not plugin.active:
--- a/pyblish/logic.py
+++ b/pyblish/logic.py
@@ -355,9 +355,9 @@ def Iterator(plugins, context, state=None, targets=None):
         "ordersWithError": set()
     }
 
-    # We'll add "default" target if no targets are registered. This happens
-    # when running the Iterator directly without registering any targets.
-    targets = targets or registered_targets() or ["default"]
+    # Include "default" target when no targets are requested.
+    targets = targets or ["default"]
+    targets += registered_targets()
 
     plugins = plugins_by_targets(plugins, targets)
 
--- a/pyblish/util.py
+++ b/pyblish/util.py
@@ -120,9 +120,9 @@ def _convenience_iter(context=None, plugins=None, targets=None, order=None):
 
     # First pass, collection
     tasks_processed_count = 1
-    for Plugin, instance in logic.Iterator(
-        collectors, context, targets=targets
-    ):
+    for Plugin, instance in logic.Iterator(collectors,
+                                           context,
+                                           targets=targets):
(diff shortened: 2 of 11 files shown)


pyblish/logic.py
```diff
@@ -354,9 +355,9 @@ def Iterator(plugins, context, state=None):
         "ordersWithError": set()
     }
 
-    # We'll add "default" target if no targets are registered. This happens
-    # when running the Iterator directly without registering any targets.
-    targets = registered_targets() or ["default"]
+    # Include "default" target when no targets are requested.
+    targets = targets or ["default"]
+    targets += registered_targets()
```
BigRoy: Will this change the behavior to always including previously registered targets as opposed to the `targets` argument being explicit overrides to the `targets`? Before this made sense, since explicitly providing a new list of targets would override the registered targets, but now it won't...

Shouldn't the behavior be to only use whatever is passed along when something is passed along, and if `None` then to use the registered targets? That way, if you'd want to explicitly run without the currently registered targets you still can... and if you'd want to run *with* the currently registered targets you can do so explicitly:

```python
registered_targets = api.registered_targets()
pyblish.util.publish(targets=registered_targets + ["bar"])
```

Other than that that I'm all for the fix on it changing the `registered_targets`. Nice job!

buddly27: Right, this will be an issue indeed. Instead it should be something like:

```python
# Include "default" target when no targets are requested.
_targets = targets or ["default"]

if targets is None:
    _targets += registered_targets()

plugins = plugins_by_targets(plugins, _targets)
```

I should also add a unit test to ensure that registered targets are not used when explicit targets are passed! Will work on it

Thanks for the feedback!

buddly27: Code updated, I also updated the [test_publishing_explicit_targets_with_global](https://github.com/buddly27/pyblish-base/blob/f97bcd3893e60f3b3a51659590f729989cfc8ebe/tests/test_util.py#L259) unit test to ensure that explicit targets override registered targets.

pyblish/logic.py
```diff
@@ -354,11 +355,13 @@ def Iterator(plugins, context, state=None):
...
 
-    # We'll add "default" target if no targets are registered. This happens
-    # when running the Iterator directly without registering any targets.
-    targets = registered_targets() or ["default"]
+    # Include "default" target when no targets are requested.
+    _targets = targets or ["default"]
 
-    plugins = plugins_by_targets(plugins, targets)
+    if targets is None:
+        _targets += registered_targets()
```
mottosso: This is interesting. Did anything change here? Under the two possible circumstances of either (a) passing `targets` as an arguments or (b) not passing it, in which case it's `None`, it seems the result is the same?

pyblish/util.py
```diff
@@ -123,11 +116,13 @@ def _convenience_iter(context=None, plugins=None, targets=None, order=None):
...
     # each of which may differ between task.
-    task_count = len(list(logic.Iterator(plugins, context)))
+    task_count = len(list(logic.Iterator(plugins, context, targets=targets)))
 
     # First pass, collection
     tasks_processed_count = 1
-    for Plugin, instance in logic.Iterator(collectors, context):
+    for Plugin, instance in logic.Iterator(
+        collectors, context, targets=targets
+    ):
```
mottosso: Stylistic note, please make this..

```py
    for Plugin, instance in logic.Iterator(collectors,
                                           context,
                                           targets=targets):
        ...
```

Alternatively..

```py
it = logic.Iterator(collectors, context, targets=targets)
for Plugin, instance in it:
   ...
```

tests/pre13/test_logic.py
```diff
@@ -211,6 +211,7 @@ def process(self, context):
     assert_equals(count["#"], 1)
 
 
+@with_setup(lib.setup_empty, lib.teardown)
```
mottosso: How come these decorators were added to the tests? Preferably they would only be assigned to those that need it. This test for example doesn't use the plug-in path at all, it's being told to run a specific plug-in directly.

mottosso (changes requested):
Taking another look at this, and my initial doubts was in changing `logic.Iterator` by adding the `targets` argument. But I think that's the way to go. Before, it kept up to date with tagets by looking globally, but that was no good. Now it's being handed targets directly, which I think is right.

Have a look at my minor notes, and then I think we can merge this. Good job.

pyblish/logic.py
buddly27: Not sure what you mean, `targets` are use by default if passed explicitly, otherwise "default" and registered targets would be used. Previously, registered targets were used by default, otherwise "default" target would be used. 

As [pyblish.util._convenience_iter](https://github.com/pyblish/pyblish-base/blob/acf97ef9fdc2fc506daac7817db99073a79a7c56/pyblish/util.py#L98) was automatically registering the "default" target when no explicit targets were passed, I didn't keep the exclusive condition between registered targets and "default" target.

I updated the logic to make it clearer:
```python
if not targets:
    targets = ["default"] + registered_targets()
```

pyblish/util.py
buddly27: Updated

tests/pre13/test_logic.py
buddly27: I'm a bit confused, most unit tests were failing without adding these decorators after my changes and I thought it was because these tests weren't protected against elements globally registered by other unit tests. 

But it must have been something wrong with my environment as I couldn't reproduce the error. I reverted these modifications...

pyblish/logic.py
mottosso: Oh, yes. You're right. Sorry I'm not sure what I meant either. xD

tests/pre13/test_logic.py
mottosso: The decorator is allowing tests to affect global state, by reverting any change they make once finished. But the tests without it weren't affecting global state and thus doesn't need them. For readability, I would only add them where needed such that the reader knows that if there *is* a decorator, then the test does affect global state.

mottosso (approved):

mottosso:
Thanks @buddly27!

PR #366, opened 2020-03-20
Remove 'vendorised' libraries from sys path

### LINKED ISSUE #365: Vendor libraries injected in sys.path are messing up environment
Is this line absolutely necessary?

https://github.com/pyblish/pyblish-base/blob/841f40821507964389adac01683ac86f745bb5fe/pyblish/vendor/__init__.py#L5

Because of this, if you have a tool using `click`, it picks up your vendor library which is a bit old.. You seems to be using "vendorised" libraries from the `vendor` module, so why adding them to `sys.path`?

https://github.com/pyblish/pyblish-base/blob/8588808a0e7e6f819e618909359502c5c8327e2e/pyblish/cli.py#L30

https://github.com/pyblish/pyblish-base/blob/841f40821507964389adac01683ac86f745bb5fe/run_testsuite.py#L10

buddly27:
This is not required anymore as these libraries are all imported via the
'vendor' module. It also prevent these libraries to interfere with the
environment.

This resolves #365

### DIFF
--- a/pyblish/vendor/__init__.py
+++ b/pyblish/vendor/__init__.py
@@ -1,5 +1,0 @@
-import os
-import sys
-
-# Append vendor plugins to PYTHONPATH
-sys.path.insert(0, os.path.dirname(__file__))
--- a/pyblish/version.py
+++ b/pyblish/version.py
@@ -1,7 +1,7 @@
 
 VERSION_MAJOR = 1
 VERSION_MINOR = 8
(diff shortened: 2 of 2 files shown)


mottosso:
Looks like we're good to go, thanks @buddly27!

mottosso:
Release is under way, looks like our CI system (Travis) is down with corona.

image

PR #367, opened 2020-04-01
Support loading unicode contained plugin in Python 3

davidlatwe:
### What's changed

Some of our plugins contain Unicode characters in plugin's `label` attribute or doc string. Those plugins which **registered in plugin path** can be discovered via Pyblish in Python 2 but fails in Python 3 due to bytes encoding issue.

The solution was simple, just change the `open` mode into `"rb"` in `api.discover`.

### Example

1. Save a plugin with Unicode character

```python
# /path/to/plugin/unicode_plugin.py

import pyblish.api

class UnicodePlugin(pyblish.api.InstancePlugin):

    label = "🤨"  # Unicode emoji

    def process(self, instance):
        return True

```

2. In Python 3, register the dir path and run `pyblish.api.discover` to see the test result

```python
import pyblish.api

# Remove other plugins for testing

# Register our only Unicode contained plugin
pyblish.api.register_plugin_path("/path/to/plugin")

# AssertionError will raised without this PR
assert len(pyblish.api.discover()) == 1

```

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -1337,7 +1337,7 @@ def discover(type=None, regex=None, paths=None):
             module.__file__ = abspath
 
             try:
-                with open(abspath) as f:
+                with open(abspath, "rb") as f:
                     six.exec_(f.read(), module.__dict__)
 
                 # Store reference to original module, to avoid
--- a/pyblish/version.py
+++ b/pyblish/version.py
@@ -1,7 +1,7 @@
 
 VERSION_MAJOR = 1
 VERSION_MINOR = 8
-VERSION_PATCH = 3
+VERSION_PATCH = 4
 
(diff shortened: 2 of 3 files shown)


BigRoy:
@davidlatwe could it be worth it potentially adding a test for this too or would it be overly verbose? Other than that, nice and clean solution!

davidlatwe:
Yeah, I was thinking about adding a test case for this, too ! But maybe tomorrow, I am just about to go out now. ☺️

mottosso:
Nice one. :) Yes, test would be good, just an additional file with lots of unicode should suffice. And if all existing tests pass already then that's already a giant confirmation it works.

davidlatwe:
Test case added !

mottosso:
Looks good, nice work. :)

PR #369, opened 2020-04-18
Extend plugin sorting logic

### LINKED ISSUE #368: Ensure that ContextPlugins are sorted before InstancePlugins for similar order
In most cases, it seems that the convention is to use **ContextPlugin** for the Collection stage and **InstancePlugin** for the following stages. However there might be some exception, for instance the InstancePlugin could be used in the Collection stage to initialize configuration of newly created instances.

For instance:

```python
class Collect(pyblish.api.ContextPlugin):
    """Context plugin to create dummy instances."""

    label = "Create dummy instances"

    def process(self, context):
        context.create_instance("Foo")
        context.create_instance("Bar")

class Construct(pyblish.api.InstancePlugin):
    """Instance plugin to initiate data for each instances."""

    order = pyblish.api.CollectorOrder + 0.1
    label = "Initiate default data"

    def process(self, instance):
        instance.data.update({...})
```

The `order` attribute is important to ensure that the InstancePlugin is always picked up after the ContextPlugin. But thinking about it more, I feel like it never really makes sense to have an InstancePlugin before a ContextPlugin, so I was wondering if there could be a more elegant solution to ensure that for plugins with the same order, ContextPlugin would always be sorted before InstancePlugin.

I thought about extending the [sorting logic](https://github.com/pyblish/pyblish-base/blob/3290fd5b51b496d431cc2ca608639e21e727ccbd/pyblish/plugin.py#L1509) as follows:

```python
def sort(plugins):
    """Sort `plugins` in-place."""
    if not isinstance(plugins, list):
        raise TypeError("plugins must be of type list")

    def _sort_by_type(plugin):
        """Return 1 if plugin operates on instances, 0 otherwise."""
        return 1 if plugin.__instanceEnabled__ else 0

    plugins.sort(key=lambda plugin: (plugin.order, _sort_by_type(plugin)))
    return plugins

```

What do you think?

buddly27:
Ensure that ContextPlugin are sorted before InstancePlugin when plugins have the same order.

This resolves #368

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -46,16 +46,21 @@ Intersection = 1 << 0
 Subset = 1 << 1
 Exact = 1 << 2
 
-# Check for duplicate plugin names. This is to preserve backwards compatility.
+# Check for duplicate plugin names. This is to preserve backwards compatibility.
 ALLOW_DUPLICATES = bool(os.getenv("PYBLISH_ALLOW_DUPLICATE_PLUGIN_NAMES"))
 
-# Check for strict data types. This is to preserve backwards compatility
+# Check for strict data types. This is to preserve backwards compatibility
 STRICT_DATATYPES = bool(os.getenv("PYBLISH_STRICT_DATATYPES"))
+
+# Sort plugins per order and per type (Context Plugin before Instance plugin).
+# This is to preserve backwards compatibility
+SORT_PER_ORDER_AND_TYPE = bool(os.getenv("PYBLISH_SORT_PER_ORDER_AND_TYPE"))
 
 # Check for early adopters.
 EARLY_ADOPTER = bool(os.getenv("PYBLISH_EARLY_ADOPTER"))
 ALLOW_DUPLICATE_PLUGINS = EARLY_ADOPTER or ALLOW_DUPLICATES
 STRICT_DATATYPES = EARLY_ADOPTER or STRICT_DATATYPES
+SORT_PER_ORDER_AND_TYPE = EARLY_ADOPTER or SORT_PER_ORDER_AND_TYPE
 
 
 class Provider():
@@ -1519,8 +1524,9 @@ def sort(plugins):
 
     *But may be overridden.
 
-    For plug-ins with the same order, ContextPlugin will always be sorted before
-    InstancePlugin.
+    If the PYBLISH_SORT_PER_ORDER_AND_TYPE or PYBLISH_EARLY_ADOPTER environment
+    variable is set to "1, ContextPlugin will always be sorted before
+    InstancePlugin for the same order.
 
     Arguments:
         plugins (list): Plug-ins to sort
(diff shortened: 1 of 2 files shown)


mottosso:
Thanks for this, it does look fine as far as tests go, which is a plus.

The only thing left to consider is whether it will make debugging more or less difficult in the future.

Having `ContextPlugin` before `InstancePlugin` isn't necessarily intuitive when compared to, say, an alphabetical order based on plug-in name.

Which reminds me, and I'd have to check, but it's possible that because plug-ins are coming from disk (e.g. `os.listdir`) in an alphabetical order they remain alphabetical once sorted with identical `order` numbers. That sorting just doesn't change them. And if so, this change would break up the alphabetical order which could be unexpected and hard to narrow down the cause of.

Here's what I think we could do. Currently, there are a few changes made to Pyblish that break compatibility in similar, subtle ways. And they've been put behind an environment variable where they can be "innocent until proven guilty", so to speak. They enable anyone to use Pyblish as though the feature was persistent, without affecting those that do not need it.

- https://github.com/pyblish/pyblish-base/blob/3290fd5b51b496d431cc2ca608639e21e727ccbd/pyblish/plugin.py#L53

If we could make this change only apply to those with the `PYBLISH_SORT_CONTEXTPLUGINS_FIRST` environment variable set, then it would be safe to merge immediately, and for anyone to evaluate. Then once it's been deemed safe, we could remove the variable and make it permanent.

Then you could also include it under the `PYBLISH_EARLY_ADOPTER` flag as well, so it applies to those who go all-in. The idea is to make each of those permanent on the next major version change, 2.0.

Let me know what you think.

buddly27:
Yeah that sounds like a good idea. I'm gonna update the PR accordingly.

> Having ContextPlugin before InstancePlugin isn't necessarily intuitive when compared to, say, an alphabetical order based on plug-in name.

Yeah it would definitely be an additional convention that might not suit everyone. We might as well add a way to register a different sorting logic as you suggested in the related issue? Would you like me to add this in the PR?

> Which reminds me, and I'd have to check, but it's possible that because plug-ins are coming from disk (e.g. os.listdir) in an alphabetical order they remain alphabetical once sorted with identical order numbers. That sorting just doesn't change them. And if so, this change would break up the alphabetical order which could be unexpected and hard to narrow down the cause of.

Hmm not sure about this logic, order resulting from [os.listdir](https://docs.python.org/3/library/os.html#os.listdir) is an artifact of the filesystem and cannot be relied upon, so you cannot really expect alphabetical order. 

Also the current order before sorting is not only dependent on os.listdir, it depends on:
- Plugin path registration order ([here](https://github.com/pyblish/pyblish-base/blob/3290fd5b51b496d431cc2ca608639e21e727ccbd/pyblish/plugin.py#L1317))
- os.listdir ([here](https://github.com/pyblish/pyblish-base/blob/3290fd5b51b496d431cc2ca608639e21e727ccbd/pyblish/plugin.py#L1322))
- order of plugins within a module ([here](https://github.com/pyblish/pyblish-base/blob/3290fd5b51b496d431cc2ca608639e21e727ccbd/pyblish/plugin.py#L1352))
- Plugin registration order ([here](https://github.com/pyblish/pyblish-base/blob/3290fd5b51b496d431cc2ca608639e21e727ccbd/pyblish/plugin.py#L1365))

buddly27:
Plugin updated as you suggested. I named the environment variable `PYBLISH_SORT_PER_ORDER_AND_TYPE` instead of `PYBLISH_SORT_CONTEXTPLUGINS_FIRST` to not mislead people by implying that the original sorting-per-order logic is removed.

PR #370, opened 2020-12-01
Add "warning" and "failedOrWarning" action types

jboisvertrodeofx:
This PR is related to a pyblish-lite PR: https://github.com/pyblish/pyblish-lite/pull/116

This basically adds support for having actions on plugins that were processed with a warning instead of an error. The existing "failed" action type being reserved for plugins that were processed with errors, a new "warning" action type has been added for those processed with warnings, and another "failedOrWarning" was added to handle both cases in a single action.

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -364,6 +364,8 @@ class MetaAction(type):
                           "notProcessed",
                           "processed",
                           "failed",
+                          "warning",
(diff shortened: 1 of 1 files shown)


mottosso:
Awesome :) Just gotta let those tests finish, then it's all-day merge-Tuesday.

PR #371, opened 2021-05-14
Convert emmited log messages to string

iLLiCiTiT:
## Issue
Objects passed to plugin's logger object are not converted to string in `MessageHandler` so GUI logging may not show object data at the moment of logging it.

## How to replicate bug
Put these in plugin logic:
```
my_data = [1]
self.log.debug(my_data)
my_data.append(2)
self.log.debug(my_data)
```

Run the plugin in Pyblish UI and see UI terminal output. Should see:
```
DEBUG: [1,2]
DEBUG: [1,2]
```

## Suggested solution
Convert message on all emmited records to string.

### DIFF
--- a/pyblish/lib.py
+++ b/pyblish/lib.py
@@ -50,7 +50,8 @@ class MessageHandler(logging.Handler):
 
     def emit(self, record):
         if record.name.startswith("pyblish"):
-            self.records.append(str(record))
+            record.msg = str(record.msg)
+            self.records.append(record)
 
 
 def extract_traceback(exception, fname=None):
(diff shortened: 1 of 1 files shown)


mottosso:
Thanks, but if we merge this then the next issue coming in will be "Why is my data structures converted to string during logging?" xD

This behavior is native to Python logging and nothing unique to Pyblish, and it's quite possible others depend on this behavior already so it can unfortunately not be changed as then those uses would break.

In this case, maybe subclass the plug-in and use your subclass for your custom plug-ins, with an overridden `self.log` that converts things to string?

```py
class MyContextPlugin(ContextPlugin):
  @property
  def log(self):
    return self._my_custom_logger
```

PR #372, opened 2021-10-07
add support for kwargs in instance init

hannesdelbeke:
now instance() and context.create_instance are interchangable, before it would error out since only some kwars were shared.

see forum post explaining it https://forums.pyblish.com/t/instance-vs-create-instance/658

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -770,8 +770,7 @@ class Context(AbstractEntity):
 
         """
 
-        instance = Instance(name, parent=self)
-        instance.data.update(kwargs)
+        instance = Instance(name, parent=self, **kwargs)
         return instance
 
     def __getitem__(self, item):
@@ -820,10 +819,11 @@ class Instance(AbstractEntity):
 
     """
 
-    def __init__(self, name, parent=None):
+    def __init__(self, name, parent=None, **kwargs):
         super(Instance, self).__init__(name, parent)
         self._data["family"] = "default"
         self._data["name"] = name
+        self._data.update(kwargs)
 
     def __eq__(self, other):
         return self._id == getattr(other, "id", None)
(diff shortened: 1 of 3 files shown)


hannesdelbeke:
just discovered the dev docs so PR likely will need some work
https://pyblish.gitbooks.io/developer-guide/content/implementing_a_feature.html

mottosso:
Thanks @hannesdelbeke! Yes, the main thing missing here are tests. Something to make sure it works the way you expect, and something to make sure it breaks when you expect it to. The fact that all prior tests pass is great, it means we haven't broken anything anyone depends on.

Other than that, I'm onboard with this change; normally I'd never actually use the `Instance` constructor directly, and would always go through `create_instance` but it doesn't hurt having it accessible as well.

hannesdelbeke:
cheers, will look into when I have time.
Instance is great to use directly when not making context plugins. ex child plugins, parented to other plugins

hannesdelbeke:
>  I'd never actually use the Instance constructor directly, and would always go through create_instance

probably because curently pyblish is context heavy, and parent support is still early days.
I believe QML does not show them yet, and from reading another thread the underlying infrastructure might be not in place yet.
see https://forums.pyblish.com/t/family-hierarchy-for-complex-scenes/443

mottosso:
That's true, there's a place for hierarchical instances but so far nobody has taken on the challenge. If there isn't yet an issue about this, now would be good time to open one so we can sort out the details. Such as, should the next parent continue if the child of a prior parent fails?

hannesdelbeke:
i'll probably follow up on that in an issue when i have time since it aligns with my goals for pyblish.
regarding failure and flow, ideally the user can set this based on their prefered workflow.
sometimes you want to know all the issues in your scene, sometimes you just want to know as quick as possible if there are no issues. really depends on the task you are working on.

hannesdelbeke:
i've added a test testing the new setup. 
this test would fail in the old code.
are we good to merge this?

mottosso:
Success! Great work @hannesdelbeke. :D

PR #373, opened 2021-10-08
add support to register plugins by module path 

hannesdelbeke:
**TLDR**: this PR adds support to register a plugin with the path to a python file(module). allowing you to register individual files without having to change your folder structure.

**Description**:

atm pyblish only supports registering a plugin either 
- by registering the plugin class instance
- or registering the folder containing the python files (modules) which contain the plugins

the paths are also saved in registered paths, and are returned by
- registered_paths
- _registered_paths
- plugin_paths()

i chose to reuse this variable since it requires minimum changes to the codebase, and is a lot cleaner than having a second var.
it also reads nice, registered paths contains all registered paths. both file and folder.

this was tested with pyblish QML and lite.

regarding potentially breaking old pipelines/ pyblish setups.
since before a file path was not allowed in registered paths. no one will have any workflows with files in the registered plugin paths.
this leads to my assumption it will be backwards compatible.

the main function using plugin_paths in the pyblish API is discover. which now handles filepaths correctly.
the CLI does not require changes i believe. 

incase there is a test that checks if registered paths does not contain a filepath, this test will now fail.

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -1354,27 +1354,11 @@ def discover(type=None, regex=None, paths=None):
                 log.debug("Skipped: \"%s\" (%s)", mod_name, err)
                 continue
 
-            for plugin in plugins_from_module(module):
-                if not ALLOW_DUPLICATES and plugin.__name__ in plugin_names:
-                    log.debug("Duplicate plug-in found: %s", plugin)
-                    continue
-
-                plugin_names.append(plugin.__name__)
-
-                plugin.__module__ = module.__file__
-                key = "{0}.{1}".format(plugin.__module__, plugin.__name__)
-                plugins[key] = plugin
+            _register_plugins_helper(plugins_from_module(module), plugin_names, plugins, module=module)
 
     # Include plug-ins from registration.
     # Directly registered plug-ins take precedence.
-    for plugin in registered_plugins():
-        if not ALLOW_DUPLICATES and plugin.__name__ in plugin_names:
-            log.debug("Duplicate plug-in found: %s", plugin)
-            continue
-
-        plugin_names.append(plugin.__name__)
-
-        plugins[plugin.__name__] = plugin
+    _register_plugins_helper(registered_plugins(), plugin_names, plugins, module=None)
 
     plugins = list(plugins.values())
     sort(plugins)  # In-place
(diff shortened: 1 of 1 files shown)


hannesdelbeke:
just discovered the dev docs so PR likely will need some work
https://pyblish.gitbooks.io/developer-guide/content/implementing_a_feature.html

mottosso:
I've thought about this and can see the use for it.

However, the `discover` function is at the root of *all* of Pyblish, and will need tests to account for all possible cases. All of them. On top of that, this change introduces a lot of new complexity to this function which means a ton of room for error. I'm onboard with the idea, but the implementation is currently too complex. I can take a closer look to see whether there is a simpler way of achieving it, but it may take a while. If on the other hand you can think of an alternative, simpler approach then that would be great.

hannesdelbeke:
fair enough, I'm aware discover is a core func and backwards compatibility is key.

regarding complexity of this PR:
if you look at the PR commits instead of at all changed code, it's easier to follow what's going on

the first 2 commits touch a lot of code, but do not change ANY functionality. just a refactor to avoid duplicate code.
they move code to helper functions, to remove duplicate code.

this is in preparation for the 3rd commit which introduces the actual new functionality:
https://github.com/pyblish/pyblish-base/pull/373/commits/36d9902732491e0bd240a9bdc8ee17e298098869
when only looking at this commit is the implementation still to complex?

hannesdelbeke:
would it be worth splitting this PR so we can tackle the changes separately?

[commit 1](https://github.com/pyblish/pyblish-base/pull/373/commits/6e1a1bfbe7c46d66ff5785b316f527b7625979ef) removes duplicate code
pyblish has 2 functions with the same code.
this commit moves the code to a helper function. no changes in any functionality

[commit 2](https://github.com/pyblish/pyblish-base/pull/373/commits/9ad7b9f2b6cee907c915bc56083a1d65ca3fd5ce) move the code for discovering a module to it's own function, so it's not embedded in the discover function.

notice it does introduce a small change!
instead of passing path, fname, and then later construct abspath
we now pass abspath, and later construct path, fname
this is likely the place where anything would break if it breaks.

[commit 3](https://github.com/pyblish/pyblish-base/pull/373/commits/36d9902732491e0bd240a9bdc8ee17e298098869) is where we add the new functionallity.
what happens here is that registered paths now also contains the paths of .py files, instead of folders only.
I can't think of anything where it would break backwards comp by introducing this change though.

Issue #374, opened 2021-10-11
pyblish silent fail confusing


hannesdelbeke:
when a plugin module fails criteria during the discovery phase, it fails silently resulting in confusing the developer.
is this the best way to handle this, or can we add a log error to the log.

hannesdelbeke:
more indepth example:

for plugins, any importing is usually done inside the process method for dcc specific code, so QML can access the plugins outside the dcc and wont crash when pymel is not found
example:
`import pymel.core as pm`

but if you want to import an action in a plugin, instead of hardcoding the action in your plugin module.
then we have to import at the top of the file.
If this import fails, the plugin discovery will fail silently and the plugin wont show. leaving the user guessing what happened.


**example**
`PATH = C:\my_folder\`

plugins live in 
`C:\my_folder\pyblish_plugins`

let's say we have the files (+ any __init__ needed)
`C:\my_folder\pyblish_plugins\plugin.py`
`C:\my_folder\pyblish_plugins\action.py`

when we import the action in the plugin
`import action` will fail
but `import pyblish_plugins.action` will succeed
this makes sense since pyblish is not an environment manager, and we never added the pyblish_plugins folder to the path.
however pyblish is able to import the plugin, which is in the same folder, because we registered the folder
`pyblish.api.register_plugin_path(r"C:\my_folder\pyblish_plugins")`

this results in a quite confusing case where pyblish silently fails, and it is quite tricky to debug why it fails.
when enabling DEBUG log pyblish logs "module action not found"
but action is in the same folder as plugin, which is found.
only after thoroughly investigating how it all hooks up can you find out why this happens.

i'm not sure if there is a good solution for this.
i can get around it and change the import in the plugin.
but any issues with a broken module for any other reasons wont be feedbacked to the user

mottosso:
Agreed, this has been a problem since day one. The problem has been *where* to make the error visible, as it isn't something the end-user can do anything with. So it's a developer-only error, which means we shouldn't display it in any GUI but could display in a terminal or e.g. Maya's Output Window. That said, having it more visible - even to users - wouldn't be the end of the world either.

If you have any ideas, I'd be happy to consider them.

hannesdelbeke:
when any tool or script in maya crashes it just errors out in the console (script editor, not output window)
i know it's a dev thing, but ideally you don't release broken code.
and hiding the plugin makes it more likely you'll release broken code

ex. 
enduser doesnt care if a plugin load was skipped because it's a private one started with _
[...]

PR #380, opened 2022-01-15
clear up confusion around pyblish silent fail

hannesdelbeke:
add log for plugin discovery steps
change error from debug to error

see https://github.com/pyblish/pyblish-base/issues/374


hannesdelbeke:
not yet all log.debug are tested, should trigger them all since one of them had a typo in the formatting

mottosso:
Looks good to me. And yes, at least one test to at least excersize them all would be fab.

hannesdelbeke:
tested all log.debugs in the maya console and none of them have typos. should be good to merge in.

mottosso:
You tested it.. manually.. this one time? 😄 Hm, that's quite cheeky, and wouldn't account for anyone making changes to these logs in the future. You sure you're happy with that?

mottosso:
One of the problem I have with this approach is that, yes, we've put log statements on each Python statement. But are all of them actually called? What are the situations where each of them are called? I bet many if not most of those log statements *never* happen. Which makes them bloat. But, since they are in there, nobody will ever remove them. Which makes them the worse kind of bloat.

What we should do, is handle the cases you've experienced so far. And explicitly write tests for them, to ensure that that case is given an appropriate message. If you haven't had a look at what the Pyblish tests look like yet, here's a good example.

- https://github.com/pyblish/pyblish-base/blob/master/tests/test_plugin.py#L87

That's it. A few lines in a function. So long as it's prefixed with `test_` it'll automatically get picked up during PRs such as this one, and let anyone know when/if it's no longer true.

hannesdelbeke:
that sounds like a good approach

Can  you give a bit more info on the test setup?
ex. there are 3 files named test_plugins. assume some of them are related to legacy code? 

PS: All logs have triggered, so none of them are bloat. But your point is still valid.

mottosso:
The tests are called by a runner called "nose", it's an old-school library that doesn't do much else than call any file prefixed `test_*.py` and function prefixed with `test_`. If the function fails, the test fails.

So a new test(s) in any of the existing files should suffice.

hannesdelbeke:
i've added tests for each possibility

mottosso:
Thanks @hannesdelbeke, I feel bad for asking this since you've done so much work already.. 😄 But the pre11 folder is for testing plug-ins written for Pyblish 1.1 and prior. Would it be too much to ask that those are just moved out of that folder, and into a new `plugins` folder inside of the `tests/` directory directly? Other than that, this looks great and you've checked all the boxes on my end.

hannesdelbeke:
yes saw your gitter message and was about to ask if they needed moving 👍

hannesdelbeke:
all ready for merge :)

hannesdelbeke:
PS: while working on this i noticed discover and registering a plugin directly have a lot of overlapping code. 
with similar error/debug messages.
might be worth unifying this somehow. but that's a refactor for another PR

mottosso:
Brilliant, thanks for suffering through my nitpicks! 🥳

PR #384, opened 2022-01-16
add pathlib support to deregister_plugin_path

hannesdelbeke:
@BigRoy correctly pointed out that while we added pathlib support to register path
deregister path doesn't yet support it.

this PR adds that support, and adds a new unittest


hannesdelbeke:
oh and the reason for those helper function:
i'm reusing them for more tests on non-pathlib path_registration in another branch

mottosso:
Looks good!

PR #386, opened 2022-01-17
refactor plugin validation

hannesdelbeke:
move duplicate code regarding plugin validation, 
from `plugins_from_module` and `register_plugin`, into `plugin_is_valid` 

refactor only no new code.

swapped
continue -> return False
raise Exception -> return False

depending on where  `plugin_is_valid` is used we then raise an exception or debug log anyway, so logic remains same.

now there's no more dupe code for plugin validation, and it's all in the place where we expect it: the `plugin_is_valid`  function

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -938,34 +938,8 @@ def register_plugin(plugin):
 
     """
 
-    if not hasattr(plugin, "__call__"):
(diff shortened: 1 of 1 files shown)


hannesdelbeke:
this should be ready for merge if we are happy with this

PR #388, opened 2022-03-31
update Plugin and Action in plugin.py

kosam0913:
Goal： to let developing pyblish plugin be more developer friendly (sort of)
Motivation: same as above~

Modifications：
1. add context and instance as class Plugin's attributes (not only a arguments)
2. append repair and other customize method into Plugin.actions during instantiation 
3. add context and instance as class Action's attribute (for further data transfer)

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -282,7 +282,7 @@ class MetaPlugin(type):
         for _f_name, _f_obj in inspect.getmembers(cls, predicate=inspect.isfunction):
 
             new_action_class = None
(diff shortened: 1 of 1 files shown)


mottosso:
Thanks for this @kosam0913, welcome to Pyblish. :)

The most important aspect of developing for Pyblish is to maintain backwards compatibility, so have a look at what AppVeyor is complaining about first of all.

Second would you be able to provide some examples of what your changes would mean? Some usage examples, along with some of the problem this addition solves. What problem should a developer be having to benefit from this change?

PR #403, opened 2025-03-17
Fix discovery for plugins with dataclasses and type annotations

antirotor:
## Problem

When plugin is using type annotations and dataclass like so:

```python
"""Test dataclass in publish plugin."""
from __future__ import annotations

from dataclasses import dataclass
from typing import ClassVar

import pyblish.api


@dataclass
class TestPublishData:
    """Test dataclass in publish plugin."""
    id: str
    name: str
    description: str
    version: str


class TestPublishPlugin(pyblish.api.ContextPlugin):
    """Test publish plugin."""
    label = "Test Publish Plugin"
    families: ClassVar[list[str]] = ["*"]

    def process(self, context: pyblish.api.Context) -> None:
        """Process the plugin."""
        _ = TestPublishData(
            id="test",
            name="Test",
            description="Test dataclass in publish plugin.",
            version="1.0.0"
        )
```

discovery will skip it because of `'NoneType' object has no attribute '__dict__'` error. The root of this is that *dataclass* is looking into `sys.modules` when it is dealing with type hints so it needs to have it available before it gets executed.

**Similar issue:** https://github.com/ynput/ayon-core/issues/1199

### DIFF
--- a/pyblish/plugin.py
+++ b/pyblish/plugin.py
@@ -1361,13 +1361,12 @@ def discover(type=None, regex=None, paths=None):
             module.__file__ = abspath
 
             try:
-                with open(abspath, "rb") as f:
-                    six.exec_(f.read(), module.__dict__)
-
                 # Store reference to original module, to avoid
                 # garbage collection from collecting it's global
                 # imports, such as `import os`.
                 sys.modules[abspath] = module
+                with open(abspath, "rb") as f:
+                    six.exec_(f.read(), module.__dict__)
 
             except Exception as err:
                 log.error("Skipped: \"%s\" (%s)", mod_name, err)


mottosso:
Looks good to me, @BigRoy I've added you to the maintainer list of this repo, feel free to merge when happy.

BigRoy:
I  must be stupid @mottosso but for the life of me I can't get the nose test runner to run the tests. I always get:

```
Ran 0 tests in 0.000s

OK
```

As such, I also can't get `run_coverage.py` because it errors on no testing data found.

Tested with:
```
python 3.9
nose 1.3.7
coverage 7.7.0
```

---

If I individually pass each test file to the `run_testsuite` like `tests/test_cli.py` etc then each does run the tests and they all seem to pass. ✅ 

Still, odd that I can't get `run_testsuite.py` to just run them all in one go.

---

And with a bit of love from @antirotor the missing flag seems to be `--traverse-namespace`.

Running this does what it should:
```
nosetests -v --traverse-namespace
```

Running however `run_testsuite.py --traverse-namespace` fails on two due to being unable to find `pyblish` module.

Which in particular is on the `test_cli` which invokes some things using `PYTHONPATH` (e.g. [here](https://github.com/pyblish/pyblish-base/blob/aeb7daf595a16d6303027d191d00a4ab4de6cd28/tests/test_cli.py#L225-L232)) env var but the `run_testsuite` only adds the local folder to `sys.path` and hence it's not on `PYTHONPATH` so I suppose that behavior is correct too.

---

It seems good to merge, even though running the test suite wasn't trivial?

BigRoy:
@mottosso any chance you can make a release? Unfortunately it's unclear for me how to do so. If it were just an **action** to run on Github I could've done so, but it seems there is none.

mottosso:
Releasing has changed so much in the past years, I also can't say for sure anymore.

I've added you to the PyPI project, you should be able to upload a release onto it, including surrounding build chain stuff like GitHub Actions etc.

https://pypi.org/manage/project/pyblish-base/collaboration/

Issue #399, opened 2024-02-23
when installing pyblish-base, tests are also installed


hannesdelbeke:
when doing `pip install pyblish-base`
it installs 2 folders
- pyblish-base
- tests

it seems the tests should not be part of this install.

PR #406, opened 2025-09-11
added exclude to setup to not copy tests when installing

quitinit:
When installing pyblish-base the tests directory is copied with it.
By excluding the tests we shrink the install size a little bit


quitinit:
closes #399

PR #408, opened 2026-02-10
Compatibility: use `pytest` instead of `nose`

antirotor:
### Problem

Using long unmaintained and abandoned [nose](https://nose.readthedocs.io/en/latest/#note-to-users) for running tests started finally breaking with python 3.11 and higher. This PR is fixing two issues:

- [X] invalid escape sequence in `iscompatible.py` breaking on python 3.12 because warnings became SyntaxErrors
- [ ] switch from`nose` to `pytest` (some cleanup still pending and dependency handling delegated to `uv`
- [ ] switch to `uv`

### DIFF
--- a/pyblish/vendor/iscompatible.py
+++ b/pyblish/vendor/iscompatible.py
@@ -2,7 +2,7 @@
 Python versioning with requirements.txt syntax
 ==============================================
 
-iscompatible v\ |version|. gives you the power of the pip requirements.txt
+iscompatible gives you the power of the pip requirements.txt
 syntax for everyday python packages, modules, classes or arbitrary
 functions.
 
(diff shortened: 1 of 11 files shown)


antirotor:
There are also two tests failing (one silently), not related most likely to switch to pytest:

tests/test_plugin::**test_argumentless_explitic_plugin** (seems like for Instance plugin, the check for arguments doesn't work?
tests/test_events.py::**test_plugin_failed_event** counting failures doesn't work

mottosso:
I've added an alternative solution here, that doesn't require a change of framework with more (many more!) dependencies.

- https://github.com/pyblish/pyblish-base/pull/409

pytest not only pulls in tons of dependencies, but some of those dependencies are compiled for the platform as well, making it difficult to vendor.

Have a look and see if it fits.

antirotor:
I've separated #410 from this as it is the real blocker.

PR #409, opened 2026-02-12
Update run_testsuite.py to support Python 3.10+

mottosso:
The old and trusty `nose` test suite needs a patch to continue working with Python 3.10 and above.


moqodow:
Hello @mottosso, 

this patch makes nose work until python 3.11. In 3.12 it does not help as `imp` required by nose is removed from python. 
Did you consider porting the tests to **nose2** as this is still actively supported.

Greetings,
Wilfried

mottosso:
nose2 would be fine yes :+1:

PR #412, opened 2026-05-28
Feature: move tests to nose2

moqodow:
Hello,

as with the comment in https://github.com/pyblish/pyblish-base/pull/409#issuecomment-4528891251 by @mottosso this PR switches the tests from nose to nose2.

Kind regards,
Wilfried


mottosso:
Ok, looks straightforward, nice work. Do you think we can make appveyor work with this as well? It would be fine to reduce the Pythons to just 1 version, and use something more recent like 3.11 or 3.12

This is what you'd need to edit:

- https://github.com/pyblish/pyblish-base/blob/master/appveyor.yml

moqodow:
> Ok, looks straightforward, nice work. Do you think we can make appveyor work with this as well? It would be fine to reduce the Pythons to just 1 version, and use something more recent like 3.11 or 3.12
> 
> This is what you'd need to edit:
> 
>     * https://github.com/pyblish/pyblish-base/blob/master/appveyor.yml

Ok. Thanks, missed this file. Will look into it and change it.

moqodow:
> > Ok, looks straightforward, nice work. Do you think we can make appveyor work with this as well? It would be fine to reduce the Pythons to just 1 version, and use something more recent like 3.11 or 3.12
> > This is what you'd need to edit:
> > ```
> > * https://github.com/pyblish/pyblish-base/blob/master/appveyor.yml
> > ```
> 
> Ok. Thanks, missed this file. Will look into it and change it.

@mottosso  I updated appvoyer but added python version 3.8 to 3.14.

The free threaded python versions are not added but work as I'm tested them locally.

mottosso:
Looks like some builds pick up an older Python, as if `pip.exe` is coming from Python 2 rather than the one from the environment.

- https://ci.appveyor.com/project/mottosso/pyblish/builds/54121436/job/erur4ums5r7koa2d

And how come there are only 178 tests running, compared to 193 from before?

- https://ci.appveyor.com/project/mottosso/pyblish/builds/53552242/job/iuwrayg6w6lni041

mottosso:
It's starting to smell like vibe coding and I'm cautious about spending much time to review this.

moqodow:
> It's starting to smell like vibe coding and I'm cautious about spending much time to review this.

Hello @mottosso, 

I'm not using any LLM or the like. I'm not a fan of them either. I just try to get the test to run on both Linux and Windows without changing them. As I don't have a Windows PC, I have to rely on the CI to see if the change does not introduce a regression. I did change the commit in place and force pushed for testing purpose only. After the testing I did do a rebase to clean up the chaotic commits.
I didn't follow up on your last comment till now as I still wanted to look into something.

The missing tests from last time where the doc tests. I didn't configure them correctly in nose2 config.
Also python2 is always included in the default CI image of appveyor. I don't know the reason why it was picked up sometimes. I had not change anything.
The mock library is included in python unittest since version 3.3 therefore changed the test to use it. 
Also the change to `uv run` is to avoid the environment picking up old vendored mock libraries that should not be present in the build environment. 

If there is a way to have a complete clean environment for each time, please point me to the documentation, I could not find any relevant reference. 

Out of curiosity, what did give the impression that this is vibe coded?

Kind regards,
Wilfried

mottosso:
Ok, I think this looks good and proper. There are a lot of edits, which makes any PR hard to review. But the edits appear mostly identical across many files and I can't spot anything off. Great work.

